PR #138, opened 2014-09-16
ChunkSize  should be size entire file in bytes minus 8 bytes

evilpete:
Ref : https://ccrma.stanford.edu/courses/422/projects/WaveFormat/


mossmann:
I cherry-picked the ChunkSize fix.  I decided not to merge the other commit because I don't want to add more RX file formats without also adding the same for TX.  Also, I'd prefer an option other than -O (for output) for something that could apply to both input and output.

Issue #87, opened 2013-06-26
multi-device support


mossmann:
libhackrf and hackrf-tools should support multiple HackRF devices connected over USB.  It should be possible to select a particular device for a given operation.  It should be straightforward to use two HackRF devices together for full-duplex operation.

hessu:
I grabbed Jared's branch and tried to polish it a bit. Feedback welcome in pull request #158.

hessu:
@dominicgs merged pull request #158 yesterday, and multiple devices support is now in the git master branch. I'll submit the gr-osmosdr changes this weekend. They're now in my feature branch: https://github.com/hessu/gr-osmosdr/tree/feature/hackrf_multidev

dominicgs:
This has been merged and in use for over a month now.  I think we can close this issue.

PR #158, opened 2015-02-23
Multiple device support #87

hessu:
This pull request contains support for multiple hackrf devices (#87). I grabbed Jared's multiple-devices branch, rebased, and did most of the TODO items listed in the commit message. It also doesn't seem to do the intermittent crashes any more. Feedback welcome.

hackrf_info can now list multiple devices, and obtain information about them even if they do not have the updated firmware (board 1 here is updated, board 0 is not):

    Found HackRF board 0:
    Firmware Version: 2014.08.1
    Part ID Number: 0xa000cb3c 0x00644746
    Serial Number: 0x00000000 0x00000000 0x12345678 0x12345678
    
    Found HackRF board 1:
    USB descriptor string: 00000000000000001234567912345679
    Firmware Version: git-1543e4a
    Part ID Number: 0xa000cb3c 0x005e4758
    Serial Number: 0x00000000 0x00000000 0x12345679 0x12345679

The libhackrf API got a set of additional functions for this: hackrf_device_list() to obtain a device list, hackrf_device_list_open() to open up one of the devices included in the list (with an index to the list), hackrf_device_list_free() to free the list. This allows support for multiple devices even if the devices run older firmware.

hackrf_spiflash, hackrf_cpldjtag and hackrf_transfer support a -d parameter to specify the device serial number. It does not need to be padded with zeros on the left, and you may also specify any short suffix of the serial number (-d 678). The first device matching the given suffix will be opened - it will not currently complain if the suffix is not unique. The -d feature requires that the board's firmware has been updated with the changes in this pull request. hackrf_info can extract the serial number after opening the device, but the -d option needs to get it from the USB device descriptor string before trying to open the device.

### DIFF
--- a/firmware/hackrf_usb/hackrf_usb.c
+++ b/firmware/hackrf_usb/hackrf_usb.c
@@ -186,25 +186,19 @@ void usb_configuration_changed(
 	}
 }
 
-int main(void) {
-	pin_setup();
-	enable_1v8_power();
-#ifdef HACKRF_ONE
-	enable_rf_power();
-#endif
-	cpu_clock_init();
-
-
-/* HACK!
-*/
+void usb_set_descriptor_by_serial_number(void)
+{
 	iap_cmd_res_t iap_cmd_res;
+	
 	/* Read IAP Serial Number Identification */
 	iap_cmd_res.cmd_param.command_code = IAP_CMD_READ_SERIAL_NO;
 	iap_cmd_call(&iap_cmd_res);
-	if(iap_cmd_res.status_res.status_ret == CMD_SUCCESS) {
+	
+	if (iap_cmd_res.status_res.status_ret == CMD_SUCCESS) {
 		usb_descriptor_string_serial_number[0] = 66;
 		usb_descriptor_string_serial_number[1] = USB_DESCRIPTOR_TYPE_STRING;
-		for(size_t i=0; i<32; i++) {
+		
+		for (size_t i=0; i<32; i++) {
 			const uint_fast8_t nibble = (iap_cmd_res.status_res.iap_result[i >> 3] >> (28 - (i & 7) * 4)) & 0xf;
 			const char c = (nibble > 9) ? ('a' + nibble - 10) : ('0' + nibble);
 			usb_descriptor_string_serial_number[2 + i * 2] = c;
@@ -214,6 +208,17 @@ int main(void) {
 		usb_descriptor_string_serial_number[0] = 2;
 		usb_descriptor_string_serial_number[1] = USB_DESCRIPTOR_TYPE_STRING;
 	}
+}
+
+int main(void) {
+	pin_setup();
+	enable_1v8_power();
+#ifdef HACKRF_ONE
+	enable_rf_power();
+#endif
+	cpu_clock_init();
+
+	usb_set_descriptor_by_serial_number();
 
 	usb_set_configuration_changed_cb(usb_configuration_changed);
 	usb_peripheral_reset();
--- a/firmware/hackrf_usb/hackrf_usb.c
+++ b/firmware/hackrf_usb/hackrf_usb.c
@@ -32,6 +32,9 @@
 #include "usb.h"
 #include "usb_standard_request.h"
 
+#include <rom_iap.h>
+#include "usb_descriptor.h"
+
 #include "usb_device.h"
 #include "usb_endpoint.h"
 #include "usb_api_board_info.h"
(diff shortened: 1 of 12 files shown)


scateu:
Awesome!

hessu:
But does it work for you?

hessu:
I'm wondering, maybe hackrf_open() API should not be broken. It'll make a bunch of developers unhappy. Maybe there should be a new hackrf_open_by_serial() variant which takes in the serial number. Now that hackrf_open() is split to a couple functions, it'd be easy.

mossmann:
Nice work!  hackrf_open_by_serial() sounds like a good idea.  I suggest taking a look at how gr-osmosdr supports multiple devices (of rtl-sdr, for example) before we commit to an API.

hessu:
With these changes it doesn't break older gr-osmosdr versions. I've also done the necessary changes for  gr-osmosdr to support the new API: https://github.com/hessu/gr-osmosdr/compare/feature/hackrf_multidev?expand=1 - I'll submit that to upstream too.

hessu:
All seems good, but it doesn't do full duplex. If an rx block is initialised first, I only get underruns. If a tx block is initialized first, I only get overruns. If I use an rtlsdr for the RX instead of a hackrf, the same script runs full duplex just fine. Suspecting a lock somewhere in the libhackrf / libusb chain.

hessu:
Well, now it does full duplex too. hackrf_init() gets called again when gr-osmocom's second block initialises, and the libusb context that was set up first got lost. Let's just use one libusb context, as libusb says it's thread safe.

kpreid:
While you're improving block startup and shutdown, could you make sure it works for **single-device half duplex (alternately using a source and sink within one process)**, too? I haven't actually gotten to the point of testing that myself, but when I reviewed the code it looked like it wouldn't work. It'd be especially convenient if creating a new source/sink killed the previous sink/source so that I don't have to make sure their lifetimes are completely non-overlapping (or, even better of course, if it did complete fake-full-duplex with simultaneous source and sink, i.e. stopping RX only when TX is in a running flowgraph and has actual samples to process).

Feel free to tell me this is out of scope (or, even better, that it already works)…

hessu:
Hmm, I'd call "out of scope", but could look at it next once this one is done, tested and out the door. It doesn't seem like it'd be working now.

hessu:
I've had plenty of fun with this one so far, appears to be working.

I've noted a bunch of examples around the web specify "hackrf=0" as settings to pick the first hackrf, even though documentation (http://sdr.osmocom.org/trac/wiki/GrOsmoSDR#HackRFSourceSink) clearly says "Use this argument without a value". For 15 out of 16 boards, "hackrf=0" will fail to find the board.

It probably comes from the rtlsdr "rtl=0" parameter which takes an index of the USB device (although indexing is a bit unreliable, on some systems the order may change on a reboot). With a setup where one board is a transmitter I really want to have the ordering right, so using a serial number instead is preferable.

One option would be to use the parameter as an index number, if the parameter is a single digit. But I don't know, maybe it's OK to break the examples which do not follow the documentation.

kpreid:
Add a special case *in gr-osmosdr only* for hackrf=0 and print a warning?

What do the other (neither rtl nor hackrf) gr-osmosdr device drivers do for identification? Consistency is good.

hessu:
gr-osmosdr:

rtl accepts both an index number, or a serial number as a parameter ("0-based device identifier OR serial number").

bladerf says "Device instance (optional). Currently, this is a 0-based device index. In the future, additional device attributes such as (a subset of)  the serial number or desired backend would be desirable attributes to consider for this argument."

Last night I adjusted the gr-osmosdr hackrf driver to accept either (1) a 0-based device index (0 to 9), or (2) a complete serial number, or a suffix of one (2 hex characters or more). That'll be consistent enough.

hewittc:
GCC 4.9.2 complains due to a couple loop iterators being declared inside the 'for' body.

```c
hackrf-multidev/host/libhackrf/src/hackrf.c: In function ‘hackrf_device_list’:
hackrf-multidev/host/libhackrf/src/hackrf.c:288:2: error: ‘for’ loop initial declarations are only allowed in C99 or C11 mode 
  for (ssize_t i=0; i<list->usb_devicecount; i++) {
  ^
hackrf-multidev/host/libhackrf/src/hackrf.c:288:2: note: use option -std=c99, -std=gnu99, -std=c11 or -std=gnu11 to compile your code
hackrf-multidev/host/libhackrf/src/hackrf.c: In function ‘hackrf_open_usb’: 
hackrf-multidev/host/libhackrf/src/hackrf.c:356:2: error: ‘for’ loop initial declarations are only allowed in C99 or C11 mode 
  for (ssize_t i=0; i<list_length; i++) {
  ^
libhackrf/src/CMakeFiles/hackrf.dir/build.make:54: recipe for target 'libhackrf/src/CMakeFiles/hackrf.dir/hackrf.c.o' failed
make[2]: *** [libhackrf/src/CMakeFiles/hackrf.dir/hackrf.c.o] Error 1
CMakeFiles/Makefile2:123: recipe for target 'libhackrf/src/CMakeFiles/hackrf.dir/all' failed
make[1]: *** [libhackrf/src/CMakeFiles/hackrf.dir/all] Error 2
Makefile:116: recipe for target 'all' failed
make: *** [all] Error 2
```

PeeJay:
The following solves the GCC 4.9 errors:
(Other than this, I can confirm it works)
```
diff --git a/host/hackrf-tools/CMakeLists.txt b/host/hackrf-tools/CMakeLists.txt
index d9f126f..261dac5 100644
--- a/host/hackrf-tools/CMakeLists.txt
+++ b/host/hackrf-tools/CMakeLists.txt
@@ -35,7 +35,7 @@ include_directories(getopt)
 add_definitions(/D _CRT_SECURE_NO_WARNINGS)
 else()
 add_definitions(-Wall)
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu90")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99")
 endif()
 
 if(NOT libhackrf_SOURCE_DIR)
diff --git a/host/libhackrf/CMakeLists.txt b/host/libhackrf/CMakeLists.txt
index 4f1ec64..0d09848 100644
--- a/host/libhackrf/CMakeLists.txt
+++ b/host/libhackrf/CMakeLists.txt
@@ -34,7 +34,7 @@ if(MSVC)
        set(THREADS_USE_PTHREADS_WIN32 true)
 else()
        add_definitions(-Wall)
-       set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu90")
+       set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99")
 
        INCLUDE(TestBigEndian)
        TEST_BIG_ENDIAN(BIGENDIAN)
```

Hoernchen:
There is a reason for gnu90: make sure people write code that is msvc-friendly beacuse msvc doesn't do c99. In Theory c89 would be right, but libusb headers use the inline keyword, which is not part of c89, but still supported by msvc - but not by gcc in c89 mode, so gnu90 does the trick.

dominicgs:
Thanks for your comments @PeeJay and @Hoernchen.  We will fix this by moving the declaration to the start of the function as we do elsewhere in the code.

I plan to test this pull request and merge it within the next few days.

hessu:
Hi, I just pushed a commit to move the declarations to the start of the function. GCC 4.9 should now be happier.

Issue #139, opened 2014-09-20
Wiki/FAQ needs better fix for "I can't seem to access my HackRF under Linux"


duanev:
The group plugdev is eschewed by several mainstream distro devs: https://bugzilla.redhat.com/show_bug.cgi?id=815093#c4  (I realize fedora is unlikely to be used for RF hacking but Arch appears to have abandoned plugdev as well).  Arch apparently has recently introduced Access Control Lists (Aug 17, 2014 https://wiki.archlinux.org/index.php/Access_Control_Lists) but they do not seem to be in wide spread use as of yet (Sept 20, 2014: filesystems are not yet mounted with the acl option).

For the time being running hackrf_info etc. and gnuradio-companion as root appears to be the easiest fix - I will not attempt to re-add plugdev to my Arch system.  However, documentation on how to use ACL to access usb devices can hopefully be added soon.

chinkey:
I belive I am having the same or a similar problem to CaptainMcCrank.  was there a solution made ? 

my target system is lubuntu 14.04 buti a runnning as roo from a serial port on my sbc.  My hackrf when pluged in comes up in /dev correctly as  /dev/hackrf-one-1-1.2 -> bus/usb/001/003

i did noticea slight change between when i plug it into my laptop and the sbc in dmesg

laptop: 
417245.660169] usb 3-3: new high-speed USB device number 10 using xhci_hcd
[417245.676669] usb 3-3: New USB device found, idVendor=1d50, idProduct=6089
[417245.676677] usb 3-3: New USB device strings: Mfr=1, Product=2, SerialNumber=0
[417245.676682] usb 3-3: Product: HackRF One
[417245.676686] usb 3-3: Manufacturer: Great Scott Gadgets

sbc: 
[ 1927.194907@0] usb 1-1.2: new high-speed USB device number 6 using dwc_otg


is this somethign to do with xhci_hcd vs dwc_otg???

any help would be amazing.

dominicgs:
I know this has been over a year for most comments on this issue, but I believe that the changes in #159 and #185 should have solved all of these issue.  Both of these fixes were included in the 2015.07.R2 release.

PR #159, opened 2015-02-26
Attempt to detect correct group for udev rules

dominicgs:
This PR tries to detect whether the system is using usb or plugdev and builds the appropriate udev rules.

This is related to issue #139 but it does nothing with systemd ACLs.


dominicgs:
This PR has an additional commit attached (the one that @ao2 and I have been discussing).

I'll close this one and split them in to two PRs.

ao2:
@dominicgs please keep me in the loop.

FYI we decided not to add ```libusb_detach_all_kernel_drivers()``` to libusb itself because it is not a complete solution (it does not cover the case of two libusb programs racing for the same resource) but it's OK to have it in a user program and libhackrf will be e good example as the device actually has two configurations.

PR #160, opened 2015-03-02
Update hackrf.c

mrbubble62:
Small change to allow build in Visual Studio 2012.
Found on HackRF.net


guruofquality:
**bump** I needed to make this change myself. I was going to make a pull request, but this pr already existed. MSVC2012 only supports c89 (which is unfortunate), but its only this one function that needs changing where the variables must to be declared first.

Issue #163, opened 2015-03-25
Detach the kernel driver automatically


ao2:
Hi,

I saw some complaints about the kernel driver interfering with libhackrf:
https://twitter.com/michaelossmann/status/579303581829988352/photo/1

There is a fairly easy solution to that, also proposed by the kernel driver developer: let libusb detach the kernel driver.

Here is a patch which can be applied with ```git am```:
http://ao2.it/tmp/0001-Automatically-detach-reattach-the-kernel-driver-if-l.patch

I only compiled-tested the patch, I don't have a hackrf myself.

if you want you can amend it to use ```true``` instead of ```1``` but I used ```1``` because that's what libusb expects.

Ciao,
   Antonio

dominicgs:
Hi Antonio,

This patch looks great, I'm in the process of spinning up a VM to test it, but I don't anticipate any problem with it.

Thanks,
  Dominic

dominicgs:
I have performed some testing on this patch.  Unfortunately, for the system that I am testing (Kali with 3.18), it does not work.

For a reason which is not yet clear, the ```libusb_set_auto_detach_kernel_driver``` function returns ```LIBUSB_SUCCESS``` but we are still unable to claim the interface or set the configuration.

If I replace the call to ```libusb_set_auto_detach_kernel_driver``` with a pair of calls to ```libusb_kernel_driver_active``` and ```libusb_detach_kernel_driver```, it works as expected.  I am not yet certain why this works when the automatic method fails.

Additionally, the ```LIBUSB_VERSION``` define is a product of the libusbx fork, so for my version it is named ```LIBUSBX_VERSION```, I assume that this will not change in future versions.

ao2:
libusb and libusbx have been merged back together (http://libusb.info), the ```LIBUSBX_API_VERSION``` name is now used just for compatibility and possibly it will be deprecated in the future:
https://github.com/libusb/libusb/blob/master/libusb/libusb.h#L146

Obviously if you have an older libusb which doesn't have ```LIBUSB_API_VERSION``` the code from my patch will not be compiled in, but I take you changed it locally to use ```LIBUSBX_API_VERSION```.

By looking at the libsub code ```libusb_set_auto_detach_kernel_driver``` is supposed to always return ```LIBUSB_SUCCESS``` on platforms which support kernel driver detaching, that function just sets a flag, the detaching is performed at claim_interface-time.
[...]

PR #185, opened 2015-06-26
Fix GitHub issue #163 - Detach the kernel driver automatically

dominicgs:
Adds detach_kernel_drivers() and set_hackrf_configuration() functions.

detach_kernel_drivers() detaches drivers from all interfaces on the active config

set_hackrf_configuration() uses detach_kernel_drivers() to detach the drivers from the interface, then changes the config and then detaches any new drivers that may have attached themselves automatically after the config change.


dominicgs:
@ao2 You had some comments last time, so please let me know your opinion on this version of the fix.

PR #188, opened 2015-07-08
Add file repeat TX mode [-R].

scateu:
Add a "-R" parameter, when TX input file gets to end, using "-R" parameter can easily do repeat TX, which is useful for testing.


vsboost:
Very handy.

PR #196, opened 2015-07-10
Add hackrf_spiflash verbose option, make default not-verbose.

jboone:


jboone:
Rescinded. My merge was eff'ed.

PR #199, opened 2015-07-14
Use Win32Thread in Visual Studio

dominicgs:
Remove the need for pthread-win32 and use Win32Thread instead

### DIFF
--- a/host/README.md
+++ b/host/README.md
@@ -80,7 +80,7 @@ Debug version:
 `make install`
 
 ###For Visual Studio 2012 x64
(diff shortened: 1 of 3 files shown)


Hoernchen:
The obvious question is: why? The whole point of just using pthreads-win32 is to have one code base that runs everywhere without any platform specific changes.

jocover:
code have a bug in set_hackrf_configuration,must use libusb_set_configuration(usb_device, 1); 
and this fork just me to test extio and UHD for hackrf with windows

PR #212, opened 2015-08-25
rad1o USB idProduct cleanups

pcfe:
- udev rule should be lowercase (https://github.com/rad1o/f1rmware/pull/29)
- add entries for rad1o storage from https://github.com/rad1o/f1rmware/blob/master/smartflash/90-rad1o-flash.rules
- Rad1o -> rad1o
- hackrf.h was NOT adjusted to lowercase cc15


schneider42:
Would be great to have this merged

dominicgs:
Sorry, this one must have slipped past me.

Is there a reason that the udev rule IDs must be lowercase?  And the python code?

If so, why not change hackrf.h to match?  Is there a convention that I am not familiar with?

schneider42:
I guess there is no reason for the Python code, except to be consistent with other hex numbers.

The udev rule seems to be case sensitive, though.

dominicgs:
Your word is good enough for me.  I'll merge it.

PR #222, opened 2015-10-27
Comment style fixup to prevent warning in c90 complier mode

skoslowski:
In ansi/c90 mode gcc warns about extra tokens at end of #endif directive [-Wendif-labels] when hackrf.h gets included. Changing the comment style gets rid off that.


rroff:
Hi,

Have you ever seen something like that? It's a bombshell you have to see! Look at it here <http://crudoslyquy.cmedianews.com/aeapilyt>

Sent from a prehistoric stone tablet, rroff@roff.us

PR #239, opened 2016-02-21
Receive streaming implementation

Sec42:
This adds a -S option to hackrf_transfer to enable receive streaming with a circular buffer of the specified size.

This makes it possible to stream data to processes which might block for short amounts of time while reading data, without loosing data. It also gives clean notifications if data is lost and never automatically exits even if the output is blocked for some time.

This is very useful when piping the output of hackrf_transfer to another process on a CPU-starved machine.


Sec42:
@mossmann The Travis build failure is because it uses (old) gcc 4.6.3 - gcc >=4.7.0 follows the C11 standard with regard to atomic operations, and compiles fine.

dominicgs:
While I agree that gcc 4.7.0 is old enough that we could rely on it for Linux users, I am unsure how Clang supports `atomic_*` and we also have a reasonable number of people using libhackrf on Windows hosts as compiled by msvc (which I believe has terrible C11 support).

schneider42:
It's green now ;)

schneider42:
I don't have Windows (TM) machine at hand to see if it compiles there. The code only gets executed if you specify the parameter.

mossmann:
Thank you for your contribution!

PR #241, opened 2016-03-12
libhackrf: Fixed segfault in hackrf_device_list

f4exb:
If there are no dev ices connected then libusb_get_device_list segfaults. This can be tested with list->devicecount. I think that if this is 0 then it is enough to return the list pointer since devicecount value should be checked by the caller.


f4exb:
Alas this does not work

PR #250, opened 2016-04-24
Updated code per PEP-8 guidelines.

jfleach:


mossmann:
Thank you very much for your contribution, but we decided to get rid of that ancient python code entirely (7e7ce3d).

PR #262, opened 2016-06-02
Replace double quote with inches

m-ahsen:
To make it github csv friendly, see https://help.github.com/articles/rendering-csv-and-tsv-data/ for details.


dominicgs:
I like this PR but it only fixes one of four instances, so the file is still not github csv friendly.

dominicgs:
This was fixed with the merging of #266

PR #283, opened 2016-09-23
make device reusable

JunfengJia:
device can be start and stop many times now, work around for the memory leak issue#281


dominicgs:
This was fixed in #246 for all transfer threads, which is functionally equivalent to this PR although a smaller change.

Thank you very much for your contribution, but as #246 has been merged, I am going to close this PR without merging.

PR #289, opened 2016-10-06
Add rad1o board id to libhackrf

muellermartin:
The rad1o is the badge of the Chaos Communication Camp 2015 (CCCamp15).

The rad1o badge contains a full-featured SDR (software defined radio) half-duplex transceiver, operating in a frequency range of about 50 MHz - 4000 MHz, and is software compatible to the HackRF.

See https://rad1o.badge.events.ccc.de for more information.

### DIFF
--- a/host/libhackrf/src/hackrf.h
+++ b/host/libhackrf/src/hackrf.h
@@ -66,6 +66,7 @@ enum hackrf_board_id {
 	BOARD_ID_JELLYBEAN  = 0,
 	BOARD_ID_JAWBREAKER = 1,
 	BOARD_ID_HACKRF_ONE = 2,
+	BOARD_ID_RAD1O = 3,
 	BOARD_ID_INVALID = 0xFF,
 };
 
--- a/host/libhackrf/src/hackrf.c
+++ b/host/libhackrf/src/hackrf.c
@@ -1606,6 +1606,9 @@ const char* ADDCALL hackrf_board_id_name(enum hackrf_board_id board_id)
 	case BOARD_ID_HACKRF_ONE:
 		return "HackRF One";
 
+	case BOARD_ID_RAD1O:
+		return "rad1o";
+
 	case BOARD_ID_INVALID:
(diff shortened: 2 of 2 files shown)


host/libhackrf/src/hackrf.h
```diff
@@ -66,6 +66,7 @@ enum hackrf_board_id {
 	BOARD_ID_JELLYBEAN  = 0,
 	BOARD_ID_JAWBREAKER = 1,
 	BOARD_ID_HACKRF_ONE = 2,
+	BOARD_ID_RAD1O = 3,
```
dominicgs: Does the rad1o firmware report a board ID of 3?  Last time I used it, I thought it was reporting 2 (HackRF One).

muellermartin: I checked it yesterday and using `3` as board ID made `hackrf_info` recognize it correctly. I thought this was already specified, because [the wiki says so](https://github.com/mossmann/hackrf/wiki/libHackRF-API#supported-board-versions).

But I also got sceptical as you asked and looked in the rad1o fork of hackrf and they definitively use board ID of 3: https://github.com/rad1o/hackrf/blob/rad1o/firmware/common/hackrf_core.h#L39. They even have an issue (rad1o/hackrf#9) for submitting it upstream ;)

AFAIK they even use both (you can choose the mode a.k.a. `hackrf` vs. `hackrf-old` on startup to be compatible with apps which do not yet support it but the HackRF).

dominicgs: I wrote that documentation on the wiki, I wonder why I didn't make sure it matched the code?

Anyway, I've merged this now.  Thank you for your contribution.

muellermartin: Cool! Thank you, too :)

schneider42: It might be that the original FW did not do that. We got our own USB PID just before camp and might have changed the board ID around that time.

PR #295, opened 2016-12-15
Added information about `HACKRF_ERROR_LIBUSB`

Dygear:
Added information about how to fix the `hackrf_open() failed: HACKRF_ERROR_LIBUSB (-1000)` error right after building the application.

### DIFF
--- a/host/README.md
+++ b/host/README.md
@@ -23,6 +23,17 @@ produce a low cost, open source software radio platform.
 
 `sudo ldconfig`
 
+If after running these commands, then doing ...
+
+`hackrf_info`
+
+results in the following error ...
+
+`Found HackRF board 0:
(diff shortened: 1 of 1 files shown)


glabmoris:
I would've expected a better fix than "reboot the computer" to justify a PR...

mossmann:
Thank you very much for your contribution, but we decided that updating the README with only one of several different causes for that error message could make it more difficult for users having one of the other similar problems.  These problems are addressed in the FAQ: [https://github.com/mossmann/hackrf/wiki/FAQ#i-cant-seem-to-access-my-hackrf-under-linux](https://github.com/mossmann/hackrf/wiki/FAQ#i-cant-seem-to-access-my-hackrf-under-linux)

PR #302, opened 2017-01-22
Multi device hardware sync

dodgymike:
Multi device hardware sync


mossmann:
What is the purpose of usb_dummy_buffer?  It looks like a big waste of RAM.

dodgymike:
Sorry that was an early attempt at pushing known data through to the host so I could filter it out. Should've been removed ages ago.

mossmann:
Cool.  That's what we figured.  Everything else looks good.

mossmann:
I merged it, but it was tricky. Can you test to make sure the hw_sync mode still works?

PR #306, opened 2017-01-24
Replace python one liner with script to build dfu files

dominicgs:
This removes a one line python command and puts it in to a script which supports both python 2 and 3.

### DIFF
--- a/firmware/dfu.py
+++ b/firmware/dfu.py
@@ -1,0 +1,7 @@
+import os.path
+import struct
+
+with open("_header.bin", "wb") as f:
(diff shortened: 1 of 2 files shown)


mossmann:
Did you mean to leave the rm commented?

dominicgs:
I did not.

schneider42:
FYI: For the rad1o we've made the whole thing stand alone in Python: https://github.com/rad1o/hackrf/blob/rad1o/firmware/tools/make-dfu.py

I don't know if it also runs under Python 3, but it does not require the dfu-util and makes the Makefile a bit easier: https://github.com/rad1o/hackrf/blob/rad1o/firmware/hackrf-common.cmake#L171

PR #310, opened 2017-01-27
Windows CI builds with Appveyor

dominicgs:


appveyor.yml
```diff
@@ -0,0 +1,27 @@
...
+  - mkdir c:\projects\hackrf\host\build
+  - cd c:\projects\hackrf\host\build
+  - cmake -G "Visual Studio 14 2015 Win64" \
+    -DLIBUSB_LIBRARIES="C:\libusb\MS64\dll\libusb-1.0.lib" \
+    -DLIBUSB_INCLUDE_DIR="C:\libusb\include\libusb-1.0" \
+    -DTHREADS_PTHREADS_INCLUDE_DIR=c:\pthreads\Pre-built.2\include \
+    -DTHREADS_PTHREADS_WIN32_LIBRARY=c:\pthreads\Pre-built.2\lib\x64\pthreadVC2.lib \
+    -DPKG_CONFIG_EXECUTABLE="C:\pkg-config\bin\pkg-config.exe" \
+    ..
+  - msbuild .\ALL_BUILD.vcxproj /logger:"C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
```
vors: - Maybe separate this into a windows build script? Multilines and yaml could play bad together for maintainability.
- would be cool to publish artifacts https://www.appveyor.com/docs/packaging-artifacts/
- I'm not 100% sure, but it could be that appveyor would not fail the build, when written in this way.

dominicgs: I don't know a lot about appveyor or building on Windows systems.  I've only recently started looking at it, so any suggestions are welcome, especially if you have experience or examples you can recommend that I look at.

Publishing artifacts is definitely on my todo list.

Appveyor definitely failed this build multiple times before I got it right, so I think it's going to give us an accurate build status, but if there are things to be improved here I'm happy to hear suggestions.

Perhaps this is a discussion for email or an issue rather than this (now merged) pull request?

vors: I think I already wrote all the things from the top of my mind.
Good to know that it actually fails. I usually wrap everything on powershell script with 

```
- ps: 
```
and it has it's own quirks.

You can cc me for appveyor-related PRs.
Super happy to see CI for windows build, that's awesome! 👍 👍 👍

PR #327, opened 2017-02-03
Device selection options

dominicgs:


dominicgs:
Commit 55d7e7f standardises on -d <serial number> for all tools (`hackrf_si5351c` supports -D <device index> for historical compatibility)

Commit feeb7a9 adds -d <serial number> support to rffc5071 and max2837 tools

PR #329, opened 2017-02-03
rad1o support

schneider42:
I'm putting this out here so you can have a look. Feel free to not merge this before your planned release (or not at all).

Few thoughts:

- There are quite a few commits and it might be a better idea to squash all of them into a single commit.
- The mixer abstraction is a bit clumsy. My goal was to not have to modify the RFFC5071 code.
- The display on the rad1o will not do anything. I'm planning to try and add support for it at a later point.

The main difference between the HackRF series of devices and the rad1o is the mixer and pin configuration. This is where most of the changes are.


schneider42:
Without testing it: I'm pretty sure https://github.com/mossmann/hackrf/commit/2163ebac9c49627692e6a4b1cf99972bab415ba7 breaks rad1o support. I'll have to take care of that before this can be merged.

mossmann:
Yes, that seems likely.  It saved 45 mA, though, so I think it is worth the trouble.

Issue #323, opened 2017-02-01
replace debug utilities


mossmann:
Create a single debug utility that replaces:
hackrf_max2837
hackrf_rffc5071
hackrf_si5351c

PR #335, opened 2017-02-08
Single debug tool

dominicgs:
This is my first run at a combined tool for reading and writing registers in the max2837, rffc5071, and si5351.


dominicgs:
This addresses issue #323

PR #351, opened 2017-02-19
fix(rad1o): Restore old clock behaviour for now.

schneider42:
Most rad1o applications use more clocks than the HackRF. I've enabled the again for now and reintroduced the original clock output change for the rad1o.


dominicgs:
My change was a fairly blunt method of cutting down power consumption by turning off all top level clocks that we weren't using.

I think it would be good to investigate this further for rad1o in the future to see if we can turn off more peripheral clocks while leaving the required base clocks on.

PR #355, opened 2017-02-28
Firmware cleanup

dominicgs:
Turn off unused clocks
Remove jellybean support (experimental dev hardware)
Reduce spurious emissions


schneider42:
Thanks for the nice rad1o integration :)

PR #399, opened 2017-08-10
Added decoupling capacitors to Operacake.

yhetti:


mossmann:
Good start!
Please add decoupling caps to supply pins on U4 and U5.
Also fix a few unconnected pads (run DRC and then switch to "Unconnected" tab).

Issue #322, opened 2017-02-01
firmware transceiver rework


mossmann:
Change the firmware to enhance TX and RX functions.  Both full-duplex (with add-on hardware) and half-duplex should be supported in the same firmware.  TX and RX should both avoid the use of stale buffers.  A callback system or other mechanism should allow special processing (e.g. sweep mode or signal source mode) in an extensible way.  The firmware should detect non-responsive USB host applications and return to an idle transceiver state.

dominicgs:
Something like this:
![img_20170127_101835 2](https://cloud.githubusercontent.com/assets/91747/22603034/f2603700-ea02-11e6-9f07-47aaa50f8c95.jpg)

miek:
Much of this was merged in #707. #180 can track further work on automatic return to idle.

PR #408, opened 2017-09-21
Firmware cleanup

dominicgs:
These commits separate and tidy some aspects of the firmware to allow for the transceiver to be reworked (issue #322).


dominicgs:
Don't merge this, it breaks sweep mode.

PR #445, opened 2018-01-12
libhackrf: refactoring and documentation

taktoa:
Still a bit of a work-in-progress, but mostly done (modulo code review).


taktoa:
Just noticed that 0a545bdd5e5263b53c553a01d53209cf9f10edee isn't signed by my GPG key for some reason, so I'm force-pushing a rebased version of the history where all commits are signed.

mossmann:
We use tabs.  Please don't submit pull requests changing tabs to spaces.  It is neighborly to comply with the whitespace and coding style preferences of the project you are contributing to.  If there is other stuff here you think we should consider, please send us a PR that is not cluttered by style changes.  Thank you!

PR #464, opened 2018-03-14
Make compatible with kitspace.org

kasbah:
Hello, I created the project sharing site [kitspace.org](https://kitspace.org) which makes it possible to buy PCBs and components with less clicks. 

I added the files that would be required to put it up on the site. I went through the CSV BOM and added all the right retailer SKUs using [an alpha version of a bom-builder tool](https://bom-builder.kitspace.org) and saved it in the currently required tab-separated format. Some part selections could probably be improved with alternatives but I wanted to be sure not to introduce any mistakes. 

Would you be happy to put up the hackrf-one on Kitspace?


mossmann:
Thank you for your contribution, but we do not want to have to maintain lists of distributor part numbers.  Our bill of materials source of truth is within the KiCad source files.  We already have difficulty keeping our bom.csv files up to date with the KiCad source even without adding this complexity.  kitspace.org looks cool!

kasbah:
Fair enough, would you be up for adding it without any retailer parts? (Though it defeats the purpose of the site a bit -- maybe one day we can automatically figure out distributor parts from MPNs). 

Have you seen [KiField](https://github.com/xesscorp/KiField) by the way?

kasbah:
Also, I could volunteer to try and keep the retailer parts fresh and in sync as much as possible.

kasbah:
Hey @mossmann , I added CSV support to Kitspace so the only change to your repo would be adding a kitspace.yaml pointing it at the right files. I have amended my branch to only be [that change](https://github.com/kitspace-forks/hackrf/commit/60dec4d4956ae895be21a3cb7826e705059292ab). Here is a [preview what the page would look like](http://add-hackrf.preview.kitspace.org/boards/github.com/kitspace-forks/hackrf/). 

Retailer parts offer still stands, though putting this up as is may also be good motivation for me to work on enabling 1-click purchased just from MPNs.

PR #465, opened 2018-03-19
Generalizing fftw build-time bits

kdunn926:
When trying to compile on a Mac OS system with FFTW installed in an odd place, I ran into some issues.  These small changes to the `hackrf-tools` cmake file seemed to sort things out.


dominicgs:
@kdunn926 Is there a reason that you closed this PR?  I thought it was good and was hoping to merge it.

kdunn926:
It seemed to break the build system. My guess is the FFTW library didn't provide a pkg-config file but not sure how the CI stuff works.

PR #469, opened 2018-03-31
Don't require a C++ compiler

arnout:
By default, CMake assumes that the project is using both C and C++.  By
explicitly passing 'C' as argument of the project() macro, we tell CMake
that only C is used, which prevents CMake from erroring out if a C++
compiler doesn't exist.


dominicgs:
Looks good to me.  Thanks for the contribution @arnout

PR #513, opened 2018-08-17
easy flash script to update firmware

ZeroChaos-:
Supports base hackrf firmware, either installed or in firmware-bin (tarball release), as well as portapack and portapack-havoc (if installed)


ZeroChaos-:
okay, this seems to work properly at this time.  I believe it eases updating hackrf firmware for "normal users" and replaces similar scripts such at these: https://github.com/sharebrained/portapack-hackrf/tree/master/host/bin

ZeroChaos-:
test failure appears unrelated

mossmann:
Thank you very much for your contribution, but I decided against merging it.

Making firmware updates easier is something we would like to do, but I don't think this script accomplishes that for most users.  It is a helper script that adds an additional method of updating firmware without eliminating the need for existing methods, so it could make our already confusing update process even more confusing.  It only works on Linux while our other host tools run on multiple platforms.  It looks for firmware files in a directory that we don't currently provide a method to install to.  It doesn't update the CPLD, which is an important part of the firmware update process.  It requires the user to put the device into DFU mode which is an unnecessary complication in most cases and requires physical access to the device.

We are very interested in improving our existing host tools for firmware update (hackrf_spiflash and hackrf_cpldjtag) to make them easier to use, possibly replacing them with a unified cross-platform binary (hackrf_firmware).  The major development efforts required to make this happen are integration of a DFU function in host code (so we don't have to call dfu-util) and improvements to the CPLD update process (e.g. eliminating the need to reset the device after an update).

I think this script may be most useful for switching between alternative firmware images that do not provide an easy-to-use USB update mechanism.  Perhaps it could be included with alternative firmware instead of being included with HackRF.

PR #515, opened 2018-08-23
Make compatible with kitspace.org (take 2)

kasbah:
Hey after #464, I added the ability to Kitspace to read csv (and xlsx, ods etc) so you would just need to add one file to have a HackRF page. You don't need to keep any additional files in sync. What do you think?

Here is a preview of the page: http://add-hackrf.preview.kitspace.org/boards/github.com/kitspace-forks/hackrf/


mossmann:
This looks like metadata that could be maintained at kitspace.org instead of in hackrf.  We are not interested in maintaining files in our repo that are specific to a service provider.  Additionally we have never recommended building boards from any published gerbers or bill of materials, always preferring to generate them from the KiCad source files.

PR #534, opened 2018-10-02
Travis OS X fixes

dominicgs:
Use the default version of xcode and remove some things that we no longer need brew to install


dominicgs:
I highly recommend using the "squash and merge" option rather than merging all four commits.

PR #546, opened 2018-11-10
HackRF Opera Cake - GPIO test mode

dominicgs:
Adds a GPIO test to `hackrf_operacake`.  You'll need to build and flash the firmware, then run `hackrf_operacake -o 24 -g`  (24 being the address of the EUT opera cake)


dominicgs:
The Travis fail is an issue with the OS X config.

PR #552, opened 2018-11-19
Fix compile with fftw3f not in system dir

w0xel:


dominicgs:
@w0xel Thanks for your contribution!

PR #580, opened 2019-01-22
The Great Pull Request of Multiple Concerns

jboone:
I was bad and didn't keep these changes on separate branches.

In this request you will find:

* Initial implementation of the PortaPack LCD user interface for HackRF operation
* Dynamic detection of PortaPack hardware, which allows a single build of HackRF firmware to operate on HackRFs with and without a PortaPack attached.
* Adding a CMake build step for libopencm3, so it does not need to be done manually.
* Fixes for poor baseband codec / CPLD / SGPIO timing.
* Addition of a CPLD SVF file output for use in PortaPack.
* An initial attempt at a Xilinx ISE Makefile for the CPLD bitstream.
* FIxes to the multi-HackRF synchronization feature recently added to the CPLD. (I think there's a better way to implement that feature, but don't have the spare cycles to take it on right now.)


firmware/CMakeLists.txt
```diff
@@ -25,5 +25,20 @@ set(CMAKE_TOOLCHAIN_FILE toolchain-arm-cortex-m.cmake)
...
+
+include(ExternalProject)
+ExternalProject_Add(libopencm3
+	SOURCE_DIR "${LIBOPENCM3}"
+	BUILD_IN_SOURCE true
+	DOWNLOAD_COMMAND ""
+	CONFIGURE_COMMAND ""
+	INSTALL_COMMAND ""
+)
+
```
dominicgs: This change causes the `cmake ..` step to fail when building from `firmware/hackrf_usb/build/`

jboone: ...and that brings up the state of the CMake stuff. I will review and propose changes.

PR #598, opened 2019-03-02
Travis-CI: Use environment variables...

jboone:
…so others can maintain nightly builds by specifying their unique values in Travis-CI.

When you merge, you will need to add these environment variables/values to the mossmann/hackrf CI settings:

REPO=greatscottgadgets/hackrf-nightly
URL=https://greatscottgadgets.github.io/hackrf-nightly

### DIFF
--- a/tools/deploy-nightly.sh
+++ b/tools/deploy-nightly.sh
@@ -3,7 +3,7 @@ PUBLICATION_BRANCH=master
 # set -x
 cd $HOME
 # Checkout the branch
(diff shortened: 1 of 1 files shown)


dominicgs:
This might be a bit pedantic, but could we change them to something like `ARTIFACT_REPO` and `ARTIFACT_URL` now that they're defined outside of the artifact context?

jboone:
@dominicgs Sure. Should we also rename PUBLICATION_BRANCH and expose as ARTEFACT_BRANCH?

jboone:
Incidentally, I don't want to switch from ARTEFACT to ARTIFACT, since you've already established a precedent the other way.

PR #600, opened 2019-03-02
CPLD SRAM loading

jboone:
This should put to bed all CPLD version issues users have dealt with over the years. At start-up, firmware will load and verify the CPLD bitstream directly into the CPLD configuration SRAM, bypassing the contents of the flash. This does not *reflash* the contents of flash, so does not affect the longevity of the CPLD. At this point forward, firmware ships with a CPLD bitstream that matches firmware expectations.

This also includes an XSVF file processing tool, which generates bitstream data as C code that is injected into the HackRF firmware, and also generates a CRC32 that can be used to check the contents of CPLD flash or SRAM.


schneider42:
Thanks :)

jboone:
@dominicgs As promised, here's a quick overview of the patch:

The Xilinx CoolRunner-II XC2C64A has internal flash which we have always used to store the HackRF CPLD bitstream. On power-up, the chip transfers the configuration stored in the flash into a set of SRAM cells that actually control the logic fabric in the CPLD. This patch uses a technique documented in the chip's BSDL and a programming document. There is a JTAG ISC command that can load a bitstream directly into the SRAM cells, replacing the contents that were loaded from configuration flash at power-up.

Whether transferring to the CPLD's flash or SRAM, the process is done one "row" at a time. Each row is accompanied by an address. For reasons unknown to me, all programming files I've seen contain a gray-coded sequence of rows, presented in the opposite endian-ness from most other JTAG data I've seen. That's why there's an array of row address values that *seem* random, but definitely aren't.

When you want to verify that a bitstream is correct (either from flash or SRAM), some of the bits you read from the device are not used/relevant and must be masked before you compare with the expected row value. Instead of storing 98 x 35 bytes worth of mask bits, I observed there are only six unique masks. So I have those six masks stored in full, plus a mask index for each row.

I've included a Python 3 tool that computes the CRC32 of a CPLD bitstream from an .xsvf programming file (generated by the Xilinx ISE tools). The firmware implements a CRC32 calculation of the CPLD flash that is reported by the hackrf_info host tool. These values can be compared to know if the CPLD flash contents match a particular version of the bitstream. (This functionality was introduced outside of this PR.)

The Python 3 CPLD tool also generates source code that contains the CPLD bitstream data. This data is written to a source code file by the firmware build system. The data is included into the firmware and is used by the SRAM load feature. So with this PR, HackRF firmware will reload the CPLD configuration SRAM at power-up with the expected CPLD bitstream. We should not have user issues relating to firmware/CPLD mismatches. Yay!

PR #589, opened 2019-02-05
Portapack and rad1o UI restructuring

dominicgs:
This supersedes #586 by adding a commit for rad1o badge.

This builds and runs on rad1o badge as a standalone firmware. 

I cannot get this to build as part of the hackrfapp in the larger rad1o firmware.  However, I believe that compatibility has been broken for a while now and is not part of this change.

PR #602, opened 2019-03-03
PortaPack and rad1o Ui restructuring, take 2

jboone:
This incorporates all changes from me and dominicgs from the prior pull request #589, plus fixes the *very important* blinky project by using weak UI setup functions. Not exactly elegant, but *done*. I think.


jboone:
Oops, Jawbreaker builds were broken because it was trying to use PortaPack UI bits. Fixed.

PR #607, opened 2019-03-13
Allow portapack and opera cake to coexist and still enable the PP UI

dominicgs:
The side effect of this is to disable GPIO mode for OC, but I2C still works.

Using both add-ons together is super unlikely, but it's nice to support it.


firmware/hackrf-common.cmake
```diff
@@ -243,6 +243,13 @@ macro(DeclareTargets)
 		COMMAND rm -f _tmp.dfu _header.bin
 	)
 
+	# Program / flash targets
+	add_custom_target(
+		${PROJECT_NAME}-flash
+		DEPENDS ${PROJECT_NAME}.bin
+		COMMAND hackrf_spiflash -Rw ${PROJECT_NAME}.bin
+	)
+
```
miek: This added flash target is a bit out of place in this commit - it could be split out into another one. Either way, it's useful & should be merged.

Issue #609, opened 2019-03-27
CPLD CRC causes configuration reload from flash


jboone:
...which means running hackrf_info causes the CPLD bitstream to regress to the contents of flash, which may not be the same as the SRAM configuration loaded at start-up from the firmware image.

Until this is addressed, I've disabled the CPLD CRC function in libhackrf until we address this. The CPLD checksum is of limited utility anyway now, because the CPLD configuration is loaded into SRAM from the firmware image at start-up. The contents of CPLD flash are somewhat irrelevant.

Sasszem:
Does this mean that we can reconfigure the CPLD to a new bitstream without flashing new firmware by writing it into the flash using `hackrf_cpld_write` and triggering a reload with `hackrf_cpld_checksum` (using a custom-complied libhackrf or just and older one)?

PR #645, opened 2019-09-26
Sync up PortaPack UI changes.

jboone:
Summary:

* Extract clock reference selection algorithm out of si5351c driver and into hackrf_core.
* Expose clock source changed UI event to both PortaPack and rad1o code.
* Separate PortaPack UI library and application code into distinct files.
* Do not expose UI structures (keep them contained within PortaPack UI implementation files).
* Macro-ize RAM/DFU/flash CMake targets.
* Make RAM and DFU targets distinct so PortaPack can run RAM firmware without presenting a serial number of "RunningFromRAM" to USB host software.
* Remove CPLD .svf file.
* Reworked .xsvf -> .c transform tool a bit.
* #ifdef-gated checksumming of CPLD, due to issue #609.
* Arbitrary CPLD code changes.


oswalpalash:
Nice catch with the commit ddec31c0a28ec1af5aec52660d94514eac2903a7 @jboone ! hackrf_sweep was being terribly slow and it was hard to pin point the cause!

Issue #650, opened 2019-10-16
HackRF freezing after multiple calls to hackrf_start_tx/hackrf_stop_tx


Matioupi:
Hello,

I'm experiencing some random rate (but highly repeatable) freezes when trying to use hackrf though the hackrf lib API and especially hackrf_start_tx and hackrf_stop_tx.

When transmitting a large payload, but calling hackrf_start_tx only once, I have not been able to reproduce the freeze, but those occurs as soon as there are multiple calls.

I set up a minimal source code example that I hope will allow to reproduce the behaviour and tell me what I did wrong (most probable), or helps spotting a deeper burried issue :-)

The freeze seems to occur because hackrf_is_streaming() API function keeps returning HACKRF_TRUE despite the fact that it should have returned for a while.

### Steps to reproduce
1. Compile the minimal example in attachment : [hackrf_tx_freeze.zip](https://github.com/mossmann/hackrf/files/3733237/hackrf_tx_freeze.zip)
 `gcc hackrf_tx_freeze.c -lhackrf -o hackrf_tx_freeze`
2. Run the example :
 `./hackrf_tx_freeze`
3. Software will start broadcasting 1Mb "0" buffer on 433MHz (low gains...) in loop. 
Wait until it freeze with output such as :
`Just before calling hackrf_stop_tx
Just after calling hackrf_stop_tx
done TX round 187
Just before calling hackrf_start_tx
Just after calling hackrf_start_tx
    end of hackrf_is_streaming() wait loop after 0.065666s
    approx tx rate : 7984162.500000 Mech/s
Just before calling hackrf_stop_tx
Just after calling hackrf_stop_tx
done TX round 188
Just before calling hackrf_start_tx
Just after calling hackrf_start_tx
    stuck in hackrf_is_streaming() wait loop for 5.000000s, TX LED is probably still ON
    stuck in hackrf_is_streaming() wait loop for 5.000000s, TX LED is probably still ON`

### Expected behaviour
If I understand correctly, the software should not spend so many time (actually forever in the hackrf_is_streaming() waiting loop) and eventually exit normally after a while.

### Actual behaviour
At some point in time, which is very random, execution get stuck in the hackrf_is_streaming() waiting loop.
At this point, the TX LED is on and will not turn OFF without a manual reset.

[...]

PR #651, opened 2019-10-16
issue #650 fix proposal

Matioupi:
Modified hackrf_stop_tx and hackrf_stop_rx to first join the transfer thread before setting the receiver to OFF mode (cf. issue #650)


Matioupi:
Only the first commit of this PR is actually related to #650, but as I suck at using git/github properly, the other commits I have made in my fork (about adding a python wrapper to libhackrf) also ended up here.

Matioupi:
The other changes than 2a089c3 may also be interesting ( python binding for libhackrf, also it's not related to related to #650 ).

mossmann:
The Python stuff is very interesting!  However, we are likely to migrate the entire HackRF software onto the GreatFET codebase which will result in a different method of access from Python.

mossmann:
I've merged just 2a089c3.  The other stuff should be considered separately.

PR #662, opened 2019-11-12
Decrease transmit latency when using hardware triggered mode.

ebarnard:
Previously, when transmit was triggered, the HackRF would exhaust its SGPIO buffers followed by the first of its USB transfer buffer before requesting more data from the USB host. If the host was using `libhackrf` this would exhaust another four USB transfer buffers before sending samples provided by the transmit callback.

Now, we fill all these buffers with user-provided data as soon as possible so that when transmit is triggered the first sample transmitted is the first user provided sample.

We also reset the HackRF's USB transfer buffer when in receive mode so the first sample send over USB is the first sample received by the device.

┆Issue is synchronized with this [Basecamp todo](https://3.basecamp.com/5295633/buckets/26082705/todos/4597425394) by [Unito](https://www.unito.io)

### DIFF
--- a/firmware/hackrf_usb/hackrf_usb.c
+++ b/firmware/hackrf_usb/hackrf_usb.c
@@ -196,6 +196,49 @@ static void m0_rom_to_ram() {
 	}
 }
 
+void initial_tx_buffer_fill_completed(void* user_data, unsigned int transferred) {
+	(void)user_data;
+	(void)transferred;
+
+	// Now we have filled the first half of the transfer buffer we reset the buffer offset to zero
+	// so that we will transmit from the start of the buffer and to trigger the transfer of the
+	// second half of the buffer.
+	usb_bulk_buffer_offset = 0;
+
+	// The SGPIO output is double buffered. Therefore we must write two samples to the SGPIO
+	// registers so that the value at the SGPIO pins when streaming starts is our first sample.
+	for (int i = 0; i < 2; i++) {
+		SGPIO_CLR_STATUS_1 = (1 << SGPIO_SLICE_A);
+
+		uint32_t* const p = (uint32_t*)&usb_bulk_buffer[usb_bulk_buffer_offset];
+		__asm__(
+			"ldr r0, [%[p], #0]\n\t"
+			"str r0, [%[SGPIO_REG_SS], #44]\n\t"
+			"ldr r0, [%[p], #4]\n\t"
+			"str r0, [%[SGPIO_REG_SS], #20]\n\t"
+			"ldr r0, [%[p], #8]\n\t"
+			"str r0, [%[SGPIO_REG_SS], #40]\n\t"
+			"ldr r0, [%[p], #12]\n\t"
+			"str r0, [%[SGPIO_REG_SS], #8]\n\t"
+			"ldr r0, [%[p], #16]\n\t"
+			"str r0, [%[SGPIO_REG_SS], #36]\n\t"
+			"ldr r0, [%[p], #20]\n\t"
+			"str r0, [%[SGPIO_REG_SS], #16]\n\t"
+			"ldr r0, [%[p], #24]\n\t"
+			"str r0, [%[SGPIO_REG_SS], #32]\n\t"
+			"ldr r0, [%[p], #28]\n\t"
+			"str r0, [%[SGPIO_REG_SS], #0]\n\t"
+			:
+			: [SGPIO_REG_SS] "l" (SGPIO_PORT_BASE + 0x100),
+			[p] "l" (p)
+			: "r0"
+		);
+		usb_bulk_buffer_offset += 32;
+	}
+
+	baseband_streaming_enable(&sgpio_config);
+}
+
 int main(void) {
 	bool operacake_allow_gpio;
 	pin_setup();
@@ -264,6 +307,30 @@ int main(void) {
 		if (start_sweep_mode) {
 			start_sweep_mode = false;
 			sweep_mode();
+		}
+
+		if (should_prepare_buffer_for_tx_or_rx()) {
+			if (transceiver_mode() == TRANSCEIVER_MODE_RX) {
+				// In receive mode we reset buffer positions so the first samples we receive are
+				// captured samples and not stale ones already in the buffer.
+				// As the SGPIO is double buffered it will produce two buffers worth of junk before
+				// the first captured sample.
+				usb_bulk_buffer_offset = 0x4000 - 64;
+				phase = 1;
+				baseband_streaming_enable(&sgpio_config);
+			} else if (transceiver_mode() == TRANSCEIVER_MODE_TX) {
+				// In transmit mode we fill the buffers over USB before we start streaming so that
+				// we only transmit user provided samples. This is especially useful in hardware
+				// triggered mode so that we transmit valid samples as soon as the trigger occurs.
+				usb_bulk_buffer_offset = 0x4000;
+				phase = 0;
+				usb_transfer_schedule_block(
+					&usb_endpoint_bulk_out,
+					&usb_bulk_buffer[0x0000],
+					0x4000,
+					initial_tx_buffer_fill_completed, NULL
+					);
+			}
 		}
 
 		// Set up IN transfer of buffer 0.
--- a/host/libhackrf/src/hackrf.c
+++ b/host/libhackrf/src/hackrf.c
@@ -259,11 +259,39 @@ static int prepare_transfers(
 			device->transfers[transfer_index]->endpoint = endpoint_address;
 			device->transfers[transfer_index]->callback = callback;
 
-			error = libusb_submit_transfer(device->transfers[transfer_index]);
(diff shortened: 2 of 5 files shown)


ebarnard:
Not sure how this will perform at high sample rates. I have only been running at 4 Msps.

(At least) the SGPIO interrupt really wants to be handled by the m0 core.

ebarnard:
@mossmann I've rebased this atop the new M0 code and simplified things a bit. It should no longer have issues at high sample rates as I'm assuming with the SGPIO shoveling running on the M0 core that it can never stall.

I'm using this to trigger transmit as close as possible to the start of an input clock signal (not the same as the 10 MHz reference clock). It looks right on a scope and seems to work with the hardware I'm testing it on. Nonetheless I'd like to test it a bit more formally.

I suppose transmitting square pulses in time with the input clock and comparing the output of the DAC to the input clock would work?

firmware/hackrf_usb/hackrf_usb.c
```diff
@@ -194,6 +194,49 @@ static void m0_rom_to_ram() {
...
+void initial_tx_buffer_fill_completed(void* user_data, unsigned int transferred) {
+	(void)user_data;
+	(void)transferred;
+
+	// Now we have filled the first half of the transfer buffer we reset the buffer offset to zero
+	// so that we will transmit from the start of the buffer and to trigger the transfer of the
+	// second half of the buffer.
+	usb_bulk_buffer_offset = 0;
+
+	// The SGPIO output is double buffered. Therefore we must write two samples to the SGPIO
```
ebarnard: I think this is the correct but could do with a second opinion.

firmware/hackrf_usb/hackrf_usb.c
```diff
@@ -264,6 +307,30 @@ int main(void) {
 			sweep_mode();
 		}
 
+		if (should_prepare_buffer_for_tx_or_rx()) {
+			if (transceiver_mode() == TRANSCEIVER_MODE_RX) {
+				// In receive mode we reset buffer positions so the first samples we receive are
+				// captured samples and not stale ones already in the buffer.
+				// As the SGPIO is double buffered it will produce two buffers worth of junk before
+				// the first captured sample.
+				usb_bulk_buffer_offset = 0x4000 - 64;
```
ebarnard: Likewise about the second opinion.

firmware/hackrf_usb/hackrf_usb.c
miek: I don't think it does work like this: `REG` and `REG_SS` are swapped in a separate operation, writing to `REG_SS` doesn't trigger a swap.

host/libhackrf/src/hackrf.c
```diff
@@ -113,6 +118,7 @@ struct hackrf_device {
 	hackrf_sample_block_cb_fn callback;
 	volatile bool transfer_thread_started; /* volatile shared between threads (read only) */
 	pthread_t transfer_thread;
+	volatile hackrf_transfer_direction transfer_direction; /* volatile shared between threads (read only) */
```
miek: I assume this would require a libhackrf major version bump?

firmware/hackrf_usb/hackrf_usb.c
ebarnard: That makes sense. To be honest I just tried this, discovered it worked and didn't bother to think about why. But I now realise that I was measuring latency.

I had a device that outputs a square wave with a 10 second period. This was used as a hardware trigger for transmit. The HackRF was fed samples to generate a matching square wave and I compared the distance between the two edges. However, I did it on the rising edge after the rising edge that triggered transmit. So I wouldn't have seen if the initial samples were wrong.

host/libhackrf/src/hackrf.c
ebarnard: `hackrf_device` is an opaque struct that is allocated by libhackrf and a pointer returned to the user via `hackrf_open(hackrf_device** device)`. This shouldn't change the ABI or require a version bump.

martinling:
Hi @ebarnard,

It's been on my todo list for a while to have another look at this PR, which I'm conscious has sat open for a long time.

Since you wrote this, quite a lot has changed in the code, and I think that most of the problems addressed here have now been fixed separately, in particular the following:

- As I think you're aware from discussion on #703, the old buffer data appearing at the start of TX/RX was fixed in #707 by @miek. That change included pre-filling the host-side buffers, thereby removing the extra TX latency of waiting for these to be emptied.

- More recently, I completed a major rewrite of the firmware's buffer management in #982. One of the things I removed in the process was the 16KB of zeros being sent out by the firmware at TX startup. The M0 is now buffer-level-aware, and will send zeros only while the buffer is empty. So there should now be no added TX latency from the firmware side, so long as the host has filled the buffer before the hardware trigger is asserted.

There is still one last line of buffering that could be optimised though, that being the SGPIO shadow registers. At the moment, the firmware does not do anything special to pre-fill these when waiting for a hardware trigger, so there will be a 32-sample delay before the first sample in the device-side buffer is outputted from the DACs.

I see that you did try to address that in this PR, but as @miek [noted](https://github.com/greatscottgadgets/hackrf/pull/662#discussion_r378377274), writing to `SGPIO_REG_SS` twice would not have had the necessary effect. I think the correct way to do this would be by writing the first 32 bytes to `SGPIO_REG`, and the second 32 bytes to `SGPIO_REG_SS`. This would need to be done before the M0 starts its normal SGPIO interrupt handling, so some thought would be needed about where to put that operation within the current architecture.

If you're interested in updating this PR in light of the above then I'd be very happy to help, in particular with advising on how to fit things in with the recent changes.

If we don't hear from you, I expect we will close this PR after a month or so as it is long out of sync with the current code. The contribution is appreciated though, and has helped to fix these issues via separate changes.

martinling:
Hi @ebarnard, thanks for following up.

> In TX mode`hackrf_transfer` looks like it still transmits four buffers of zeros before sending user data as `libusb_submit_transfer` is called for each of the four transfers before the first call to `device->callback`.

Yes, I actually ran into this myself just the other day. When I wrote up my previous summary above, I had misunderstood the effect of PR #707. It did fix the uninitialised data in the transfer buffers, but only by zeroing them out, not by preloading them with samples. I'm now planning to make libhackrf preload the transfer buffers by calling the TX callback for each one before initial submission, much as you did here. The code may look a bit different though, due to changes in libhackrf.

> In TX mode buffer underruns might still lose signal timing information (I haven't looked at this one carefully). In the event of a one buffer delay, the current code looks like it will `block 1 | block of zeros | block 2 | block 3` but it should transmit `block 1 | block of zeros | block 3` to keep signal timing.

Yes, the current code will lose the relationship between timing and sample numbers if there are TX underruns. It was basically designed to work as it did before, but to transmit zeros during underruns rather than repeating previous buffer contents. This behaviour has the advantage that you can have a stall in throughput and things will return to normal afterwards, albeit now shifted in time. For most use cases that is OK.

I think that trying to maintain signal timing would be problematic within the current API. If underruns are occurring, then the system is struggling to maintain USB throughput, and if you try to maintain timing, you can end up with TX stopping completely because the sample stream never catches up with the original timing. There's no way in the current structure for the device to say "hey, it's too late for those samples now, skip this many and give me the ones after that". The host is just sending a continuous stream of samples, and by the time the firmware sees them, discarding some of them can't help the USB flow catch up.

Instead, I focused on adding buffer statistics so that you can check whether underruns are occuring, and make make sure your system performs well enough that they don't happen in the first place. I also added a TX underrun limit setting, so that if you care about perfect timing, you can set things up such that when any underrun occurs, TX is aborted and you'll get an error.

To go beyond that and have recoverable underruns whilst also maintaining exact signal timing, I think we really need a different API. I have some plans in that direction which I need to write up, but it basically involves having a new transceiver mode in which you can schedule TX, RX, retuning and other operations at precise sample numbers. This mode would work differently at both the API and USB level, with TX and RX data being explicitly timestamped, and the host code getting constant feedback. This would help with a lot of advanced use cases.

ebarnard:
> I'm now planning to make libhackrf preload the transfer buffers by calling the TX callback for each one before initial submission, much as you did here. The code may look a bit different though, due to changes in libhackrf.

Excellent. My code to do this was a bit of a mess and involved the callback being invoked from two different threads which isn’t ideal.

> If underruns are occurring, then the system is struggling to maintain USB throughput, and if you try to maintain timing, you can end up with TX stopping completely because the sample stream never catches up with the original timing.

I was using the HackRF for GNSS signal simulation which is very timing sensitive but very tolerant to noise. In my case USB latency (perhaps due to variable host CPU loading) was such that sometimes buffers would arrive too late and cause the signal to lose timing. USB throughout was always sufficient for the buffers to catch up so such instances would just appear as transient noise.

> This mode would work differently at both the API and USB level, with TX and RX data being explicitly timestamped, and the host code getting constant feedback. This would help with a lot of advanced use cases.

This sounds like a much better solution. Mine was very much a case of minimal work required to get a reliably working test environment.

PR #671, opened 2019-12-09
Acrylic case

yhetti:


mossmann:
Is there a reason to have this in three different file formats?

yhetti:
The .dxf is there because it is replacing the .dxf that already existed in the repo (which was in PlasticCase_CAD/). Our preferred format is the .svg. And the PDF is in there because I know some people have trouble with SVGs so I gave another option.

PR #697, opened 2020-01-22
remove specific sample rate recommendations

mossmann:
HackRF One supports arbitrary sample rates from 2 Msps to 20 Msps.  In
early development we had ideas about preferred sample rates that we no
longer consider valid.


miek:
Looks good to me :+1:

PR #716, opened 2020-03-16
fix _FILE_OFFSET_BITS 64 position in hackrf_transfer.c 

Jakub6:
hackrf_transfer failed to send file with size over 2G (in case 32bit system). Fix in hackrf_transfer.c  to change the position of the #define _FILE_OFFSET_BITS 64, must be defined before #include<sys/stat.h>


mossmann:
Thank you for this fix!

PR #748, opened 2020-06-16
Update hackrf_sweep.c

ggatis:
change if ( fd != NULL ) to if ( ( fd != NULL ) && ( fd != stdout ) )
before fclose( fd );

### DIFF
--- a/host/hackrf-tools/src/hackrf_sweep.c
+++ b/host/hackrf-tools/src/hackrf_sweep.c
@@ -757,7 +757,7 @@ int main(int argc, char** argv) {
 		fprintf(stderr, "hackrf_exit() done\n");
 	}
 
-	if(fd != NULL) {
+	if ( ( fd != NULL ) && ( fd != stdout ) ) {
 		fclose(fd);
 		fd = NULL;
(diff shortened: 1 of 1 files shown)


miek:
Is there a problem with calling `fclose` on `stdout`? As far as I'm aware there isn't any issue with it and it's sometimes recommended, to ensure that any output is flushed before exiting.

ggatis:
Hi!
 
stdout is provided by OS and IMHO it should be managed by OS even if there
are no issuses.
Actually I lost print operator functionality in python when sweep finished
after calling it as dll.

Kind regards ~

Gatis

 


  _____

mossmann:
I'm going to accept this patch because it seems that the potential consequences of closing stdout are slightly worse than the potential consequences of not closing stdout, but we do suggest using `dup()` if you are doing things like dynamically linking another program's `main()` into your own.

mossmann:
Thank you for your contribution!

ggatis:
OK! Thank you for pointing to dup()!

Kind regards ~
Gatis

ggatis:
You’re welcome!

Issue #804, opened 2020-11-06
libusb - hackrf library is using libusb wrong, leading to stalls on (at least) freebsd


erikarn:
Hi!

I've been digging into why hackrf on gqrx/gr-osmosdr has been behaving poorly on freebsd over the last couple of years. Things work fine when RX starts, but if RX stop / RX start is then issued, the library hangs / errors out.

What's going on is:

* rx/tx stop is called
* the libusb loop thread is canceled
* pending transfers are canceled

The problem is that because the libusb loop is canceled, the pending transfers fail to be stopped because the code that marks them as canceled in libusb happens in the main loop, not in the cancel call. So, re-adding them when RX is re-started results in LIBUSB_ERROR_BUSY to be returned from libusb_transfer_setup().

So, what SHOULD happen is:

* rx/tx stop is called
* transfers need to be canceled
* then we need to wait for the callback to be called
* /then/ we can stop the libusb loop.

I'm digging into this right now to see if I can fix it up.

erikarn:
I've modified the code here to create the thread when _open() is called, and kill it when _close() is called.

That required me to separate out the transfer setup bits but that wasn't difficult.

Now stop/start/stop/start is 100% reliable.

PR #805, opened 2020-11-09
Fix libusb usage for at least freebsd around the worker thread and transfer cancellation

erikarn:
This fixes Issue #804 .

On at least freebsd-13 trying to cancel a transfer whilst the libusb thread
is not running results in the transfers not completing cancellation.
The next time they're attempted to be re-added the libusb code thinks
they're still active, and returns BUSY on the buffers.

This causes gqrx to error out when one makes DSP changes or stops/starts it.
You have to restart gqrx to fix it.

After digging into it a bit, the libusb code expects that you're actively
running the main loop in order to have some deferred actions run in the
context of said main loop thread.  This includes processing cancelled
transfers - the callbacks have to be run (if they exist) before the
buffers are properly cancelled and have their tracking metadata (a couple of
private pointers and state) removed from inside of libusb.

This patch does the following:

* separate out adding and cancelling transfers from the libusb worker thread
  create/destroy path
* create the libusb worker thread when opening the device
* destroy the libusb worker thread when closing the device
* only add and cancel transfers when starting and stopping tx/rx
* handle cancelled transfers gracefully in the USB callback

Whilst here, also make the libusb device memory zeroed by using
calloc instead of malloc.

This fixes all of the weird libusb related buffer management problems
on FreeBSD.

### DIFF
--- a/host/libhackrf/src/hackrf.c
+++ b/host/libhackrf/src/hackrf.c
@@ -119,6 +119,7 @@ struct hackrf_device {
 	void* tx_ctx;
 	volatile bool do_exit;
 	unsigned char buffer[TRANSFER_COUNT * TRANSFER_BUFFER_SIZE];
+	bool transfers_setup; /* true if the USB transfers have been setup */
 };
 
 typedef struct {
@@ -157,6 +158,8 @@ static const uint16_t hackrf_one_usb_pid = 0x6089;
 static const uint16_t rad1o_usb_pid = 0xcc15;
 static uint16_t open_devices = 0;
 
+static int create_transfer_thread(hackrf_device* device);
+
 static libusb_context* g_libusb_context = NULL;
 int last_libusb_error = LIBUSB_SUCCESS;
 
@@ -169,7 +172,7 @@ static int cancel_transfers(hackrf_device* device)
 {
 	uint32_t transfer_index;
 
-	if( device->transfers != NULL )
+	if( (device->transfers != NULL) && (device->transfers_setup == true) )
 	{
 		for(transfer_index=0; transfer_index<TRANSFER_COUNT; transfer_index++)
 		{
@@ -178,6 +181,7 @@ static int cancel_transfers(hackrf_device* device)
 				libusb_cancel_transfer(device->transfers[transfer_index]);
 			}
 		}
+		device->transfers_setup = false;
 		return HACKRF_SUCCESS;
 	} else {
 		return HACKRF_ERROR_OTHER;
@@ -269,6 +273,7 @@ static int prepare_transfers(
 				return HACKRF_ERROR_LIBUSB;
 			}
 		}
+		device->transfers_setup = true;
 		return HACKRF_SUCCESS;
 	} else {
 		// This shouldn't happen.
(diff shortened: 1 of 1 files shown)


erikarn:
(hm, it seems to work fine for stop/start but not for changing operating mode in gqrx for some odd reason, so let me go dig into it a bit more to see what else needs to be fixed with libusb use.)

erikarn:
(Update - so far the OTHER issues are inside of gr-osmosdr and their lack of clear lock / condition wait handing in the buffer management path; I'll keep poking at this and update the patch when I'm ready.)

erikarn:
Ok, this commit stack (3 commits) fixes all the libhackrf issues I've been seeing. there are hangs in gr-osmosdr because of how they currently implement receive streaming and they wait forever for buffers that may never come; I've fixed that and I'll try to upstream that asap.

erikarn:
Ugh, failing on WIndows? Ok, lemme go make it work there using Sleep() or something.

erikarn:
Ok, this passed. I'd like this patch set to be considered. thanks!

ktemkin:
It'll be a little bit before I can take a look at it -- I suspect @miek might beat me to it. :)

erikarn:
oki thanks!

erikarn:
Ping @ktemkin ! Don't suppose you could eyeball this a bit pretty please?

Thanks!


-erikarn

ktemkin:
Apologies; it's been a while and I'm only now just getting a chance to look at things.  (This potentially needs testing on multiple platforms, so it's had to wait until I had a good enough chunk of time.)

I don't have a FreeBSD box around to do testing with; so I'm going to take it on faith that you've tested enough over there and check things out using other platforms.

ktemkin:
I haven't taken a deep look at this, but at the very least this breaks API compatibility on Linux -- `hackrf_close` now fails with a return code:

```
$ hackrf_info

hackrf_info version: git-b4ea51a
libhackrf version: git-b4ea51a (0.5)
Found HackRF
Index: 0
Serial number: 000000000000000087c867dc2a8d4d5f
Firmware Version: 2018.01.1 (API:1.02)
Part ID Number: 0xa000cb3c 0x0056475f
hackrf_close() failed: unspecified error (-9999)

```

I haven't had a chance to take a deeper look, but I suspect one of the `libusb` calls needs to be handled with a bit more subtlety on Linux.

erikarn:
Ugh nooooooooooo :(

Ok, let me go and update a linux box to test this on. (amusingly all my
linux use is in VMs..) and I'll report back. libusb is .. quite frankly a
sh*t.


-adrian

erikarn:
ok, I addressed why it was throwing an error internally; no transfers were setup so checking the return value of check_transfers() failed.

Wanna try again and let me know if that works any better? thanks a bunch!

ktemkin:
Yep -- will check this out today and get back to you. :)

ktemkin:
This does seem to fix the issues on Linux; though I still need to test this macOS and Windows. :)

ktemkin:
Okays; it seems like basic tests pass over here -- and assuming the sleeps added are a workaround because other methods were non-trivial to implement cross-platform, LGTM.

I'll defer merging to @mossmann or @miek; since they're the ones who are going to have to deal with any breakage. :)

erikarn:
> Okays; it seems like basic tests pass over here -- and assuming the sleeps added are a workaround because other methods were non-trivial to implement cross-platform, LGTM.
> 
> I'll defer merging to @mossmann or @miek; since they're the ones who are going to have to deal with any breakage. :)

Thanks Kate! I look forward for further review and hopefully a merge!

erikarn:
ping! Is there anything else I can do to help get this over the line? :-) Thanks!

mossmann:
Thank you very much for this contribution!

erikarn:
Thanks for finally merging this! :-)



-adrian

PR #852, opened 2021-03-18
fix usb transfer timeout back to 1 second, limit stats reporting to 1Hz

gozu42:
the introduction of the -N option changed the sleep on the main loop in hackrf_sweep from 1sec to 50ms.

this has two side effects:
- the usb transfer timeout is 50ms now too
- and the sweep stats reporting is rather spammy when done in 50ms intervals.

this fixes both issues.


gozu42:
withdrawn, @miek has a cleaner solution for both: https://gist.github.com/miek/c178a1d3035412d1e91a1ee93709d341

Issue #871, opened 2021-04-06
hackrf_sweep: stdout binary format seems to be corrupted (file out still okay)


dmaltsiniotis:
### Steps to reproduce
1. Run the following two commands and observe the output files
2. hackrf_sweep.exe -f 100:1000 -B -w 100000 -l 16 -g 24 -1 -r out1.bin
3. hackrf_sweep.exe -f 100:1000 -B -w 100000 -l 16 -g 24 -1 > out2.bin

### Expected behavior
I expected that the out2.bin file should look _similar_ in structure and format to the out1.bin file insofar as the lengths of each binary record.

### Actual behavior
In the out2.bin file that was piped from stdout, there seem to be the occasional extra byte polluting the power value integers in the frequency bin portion of the binary data. I came across this error because a 3rd party tool I was using (QSpectrumAnalyzer) was failing to parse the binary data from the latest version of hackrf_sweep.

When specifying an actual file to output to, the binary data appears correct and consistent. When reading from standard input, the binary data is corrupted.

### Version information
**Operating system**: Windows 10 Pro

**hackrf_info output:**

### Output
Console (stderr in the case) output is the same for each command
```
C:\dev\hackrf_tools>C:/dev/hackrf_tools/hackrf_sweep.exe -f 100:1000 -B -w 100000 -l 16 -g 24 -1 -r out1.bin
call hackrf_sample_rate_set(20.000 MHz)
call hackrf_baseband_filter_bandwidth_set(15.000 MHz)
Sweeping from 100 MHz to 1000 MHz
Stop with Ctrl-C

Exiting...
Total sweeps: 1 in 0.12549 seconds (0.00 sweeps/second)
hackrf_close() done
hackrf_exit() done
exit
```

```
C:\dev\hackrf_tools>C:/dev/hackrf_tools/hackrf_sweep.exe -f 100:1000 -B -w 100000 -l 16 -g 24 -1 > out2.bin
call hackrf_sample_rate_set(20.000 MHz)
call hackrf_baseband_filter_bandwidth_set(15.000 MHz)
Sweeping from 100 MHz to 1000 MHz
Stop with Ctrl-C

Exiting...
Total sweeps: 1 in 0.18038 seconds (0.00 sweeps/second)
hackrf_close() done
hackrf_exit() done
fclose() done
exit
```

[...]

PR #872, opened 2021-04-07
Adding call to _setmode for Windows only to specify binary stdout output piping.

dmaltsiniotis:
If the binary mode flag -B is set, also set the process's stdout pipe mode to binary mode. If this is not done, the underlying c standard library functions on windows will assume text-mode and try to convert any 0x0A bytes to 0x0A 0x0D, mangling the binary output. This fixes QSpectrumAnalyzer being able to parse the binary output of the latest version hackrf_sweep on Windows.

See Issue #871


miek (approved):
I haven't got a windows setup to actually test on, but otherwise this looks good to me

mossmann:
Thank you very much for this fix!

PR #874, opened 2021-04-08
hackrf_sweep: Calculate and show sweep rate for subsecond sweeps

metayan:
With this change,
`hackrf_sweep -f 100:120 -N 100`
reports
```
Total sweeps: 100 in 0.30062 seconds (332.64 sweeps/second)
```
instead of
```
Total sweeps: 100 in 0.30061 seconds (0.00 sweeps/second)
```

### DIFF
--- a/host/hackrf-tools/src/hackrf_sweep.c
+++ b/host/hackrf-tools/src/hackrf_sweep.c
@@ -398,7 +398,7 @@ int main(int argc, char** argv) {
 	struct timeval time_now;
 	struct timeval time_prev;
 	float time_diff;
-	float sweep_rate;
+	float sweep_rate = 0;
 	unsigned int lna_gain=16, vga_gain=20;
 	uint32_t freq_min = 0;
 	uint32_t freq_max = 6000;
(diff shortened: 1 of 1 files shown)


host/hackrf-tools/src/hackrf_sweep.c
```diff
@@ -741,6 +741,8 @@ int main(int argc, char** argv) {
 
 	gettimeofday(&time_now, NULL);
 	time_diff = TimevalDiff(&time_now, &t_start);
+	if((sweep_rate == 0) && (time_diff > 0))
```
dmaltsiniotis, line 744: There is a chance that this check for ```sweep_rate == 0``` may be checking against an uninitialized instance of sweep_rate because of a preexisting issue (see #869 and #870) . I would also recommend initializing sweep_rate to 0 on line 401.

metayan: Thanks for spotting this.
Pushed a change.

miek (approved):

mossmann:
Thank you!

PR #890, opened 2021-05-19
Report amplitude once per second during receive

cjheath:
When using hackrf_transfer for receive, it's difficult to determine the correct gain settings.

This PR adds an approximate "dB fullscale" value to the once-per-second report, so you can do some manual gain control.

E.g.
```
$ hackrf_transfer -r cap1 -f 101.7e6 -s 20e6 -n 40000000 -a 1 -l 24 -g 20
call hackrf_set_sample_rate(20000000 Hz/20.000 MHz)
call hackrf_set_freq(101700000 Hz/101.700 MHz)
call hackrf_set_amp_enable(1)
samples_to_xfer 40000000/40Mio
Stop with Ctrl-C
40.1 MiB / 1.004 sec = 39.9 MiB/second, amplitude -7.6 dBfs
40.1 MiB / 1.004 sec = 39.9 MiB/second, amplitude -4.6 dBfs

Exiting... hackrf_is_streaming() result: streaming terminated (-1004)
Total time: 2.00845 s
hackrf_stop_rx() done
hackrf_close() done
hackrf_exit() done
fclose() done
exit
```

It's not a true dB(fs) but an approximation based on the average magnitude of the individual I & Q values, converted to dB relative to 128 (the maximum magnitude of a signed byte).

You can inspect the numeric I&Q values in a baseband file (except WAV) using `od -t d1`. If you see values above 100 or < -100, you're probably getting overload.


mossmann:
Thank you! This looks very useful! I would like to get @miek's opinion on the implementation before merging.

mossmann:
There is a fairly common scenario in which very brief transmissions cause clipping, so perhaps we could output both the average and peak.

cjheath:
I figure if you know you're dealing with a bursty signal, just calibrate the settings to leave more headroom.

I don't know if the ADCs clamp or wrap, but either way it's hard to know if overload is occurring. We could build a histogram from 0..127 and calculate something from that. A peak filter (biggest I or Q value in a short window) would assist in building a meaningful histogram.

Or we could add an AGC mode to the firmware (perhaps using an existing RSSI or clipping indicator, if any?), and augment the IQ data with the applied gain settings - but that would not be backward compatible.

But I figure that anything is better than nothing.

The Appveyor build is failing because its `math.h` doesn't define `M_LN10`, sigh. And I just fixed he short frame detection.

host/hackrf-tools/src/hackrf_transfer.c
```diff
@@ -390,6 +406,13 @@ int rx_callback(hackrf_transfer* transfer) {
 			}
 			bytes_to_xfer -= bytes_to_write;
 		}
+
+		// accumulate stream_amplitude:
+		for (i = 0; i < bytes_to_write; i++) {
+			int val = (signed char)transfer->buffer[i];
+			stream_amplitude += (val < 0 ? -val : val);
```
miek, line 413: This should just use `abs`

host/hackrf-tools/src/hackrf_transfer.c
```diff
@@ -1058,20 +1081,33 @@ int main(int argc, char** argv) {
...
 			time_difference = TimevalDiff(&time_now, &time_start);
 			rate = (float)byte_count_now / time_difference;
 			if (byte_count_now == 0 && hw_sync == true && hw_sync_enable != 0) {
 			    fprintf(stderr, "Waiting for sync...\n");
 			} else {
-			    fprintf(stderr, "%4.1f MiB / %5.3f sec = %4.1f MiB/second\n",
-					    (byte_count_now / 1e6f), time_difference, (rate / 1e6f) );
+			    // This is only an approximate measure, to assist getting receive levels right:
+			    double	full_scale_ratio = ((double)stream_amplitude_now / (byte_count_now ? byte_count_now : 1))/128;
+			    double	dB_full_scale_ratio = 10*log(full_scale_ratio)/M_LN10;
```
miek, line 1102: Is there a reason not to use `log10` here? That would also remove the need for the special-case defining the LN10 constant.

Also, do we need to use double precision for this?

host/hackrf-tools/src/hackrf_transfer.c
```diff
@@ -1058,20 +1081,33 @@ int main(int argc, char** argv) {
...
+			    if (dB_full_scale_ratio > 1)
+			    	dB_full_scale_ratio = nan(0);	// Guard against ridiculous reports
```
miek, line 1104: I think this should use the `NAN` constant, and that should also fix the build error on Travis

cjheath: Done. NAN wasn't listed in my man page (so I didn't know its standards compliance) and nan(0) was.

cjheath:
My apologies. <embarrassment>In haste I pushed the executable instead of the source.</embarrassment>. Fixed now

cjheath:
What's the status of this PR? Any reason not to merge it?

straithe:
Thank you for checking in! Our small team of developers is spread across all Great Scott Gadgets projects, so sometimes we can't get back to things as quickly as we'd like. Hopefully we can back to this PR this week or next.

mossmann:
Squashed (to avoid including that binary) and merged. Thank you for the contribution!

cjheath:
Excellent plan, and thanks.

Issue #528, opened 2018-10-01
Add Pseudo-Doppler support to `hackrf_operacake` tool


dominicgs:
Depends on #527 

The current pseudo-doppler mode is triggered by a special combination of opera cake port settings.  We should have a new USB API call to start and stop PD mode.

This could be an extra call from an existing tool or a new command line tool.

The tools should be documented in the way that other tools are documented on the wiki.

williamg42:
I've been working on a tool an have already merged docprofsky's firmware and the most recent hackrf master but I have seen no sign of special opera cake port settings in the firmware. However calling the appropriate USB APIs do not seem to start the antennas feeds rotating. Any information on that special combination of opera cake port settings would be helpful.

Nzbu:
> 
> 
> I've been working on a tool an have already merged docprofsky's firmware and the most recent hackrf master but I have seen no sign of special opera cake port settings in the firmware. However calling the appropriate USB APIs do not seem to start the antennas feeds rotating. Any information on that special combination of opera cake port settings would be helpful.
> 
> For anyone looking for information:
> You need to set PA and PB to 0, in the operacake_set_ports(uint8_t address, uint8_t PA, uint8_t PB) command

How did you set the PA and PB ports to 0? I've tried with the "hackrf_operacake -o 24 -a 0 -b 0", but it does not seem to start the switching. I've also merged these codes.

williamg42:
If you check my fork I had to hard code it during the hackrf startup
sequence. I was not able to get the USB commands to work either.

Nzbu:
Thank you! Is the version of the code on your fork functioning, so I could try it?

williamg42:
It should be. I don't remember what I set the "rotational rate" but you can
find that in the main code.

maxroseGIT:
Hi @williamg42, thanks for the reply on @Nzbu question.

Can you be a bit more specific: What is the 'main-code' from your point of view?

And where did you hard code PA and PB to get the antenna switching started?

Thanks for letting us know.

williamg42:
Sorry, I guess I should have phrased that better. The command is placed in
the main function in hackrf_usb.c.

maxroseGIT:
Hey @williamg42, thanks for clarification. Can you give us some hint how much time you spend figuring out all required changes to make the switching work? Did you have contact to the developers?

For me it seems quite a challenge to dig through the code and find the important places of code that make the opera-cake working like I want it to...

[...]

PR #931, opened 2021-08-11
Opera Cake SCTimer-based antenna switching

### LINKED ISSUE #527: SCT antenna switch for Pseudo-Doppler
We need to merge the changes from @docprofsky's [operacake-sctimer branch](https://github.com/docprofsky/hackrf/tree/operacake-sctimer) to the HackRF master branch firmware.

miek:
This fixes #527 and #528.

The first couple of commits add the concept of mode-switching to the Opera Cake code. This lets the user switch each Opera Cake board between manual port selections and automatic port switching over frequency ranges.

The third commit has the bulk of the new code. This is a rebased & squashed version of @docprofsky's original [operacake-sctimer branch](https://github.com/docprofsky/hackrf/commits/operacake-sctimer) with a few changes:
 * A new API for setting the port switching order and the time to dwell on each port (`operacake_sctimer_set_dwell_times`).
 * The SCTimer clock prescaler is changed from `0` to `1`, to run at half the SGPIO clock speed & be equal to the sample rate. (this may be what https://github.com/mossmann/hackrf/issues/527#issuecomment-426048548 was referring to)
 * The SCT_STATE register is used to enable/disable the SCTimer events. This makes it easy to turn off the IO pin changes when not needed, without having to mess with the overall block configuration.
 * Any USB/API additions are removed and split out into other commits.
 * Fixes a problem where one switch state would last a cycle longer than the others. This seemed to be due to the counter reset at SCT_LIMIT taking an extra cycle, so I reduced the dwell time for the first event to compensate & the timing looks good now. (ref: https://github.com/mossmann/hackrf/issues/527#issuecomment-426047531)

The final commits add the time switching mode to the CLI/API and add the API function for setting dwell times. One question I have is whether that's a good name for the API call? Another option I was thinking was `hackrf_set_operacake_time_ranges` to match `hackrf_set_operacake_freq_ranges`?

For testing, the command looks something like this:

```
$ hackrf_operacake -o 0 -m time -t A1:1000 -t A2:1000 -t A3:1000 -t A4:1000                                                                                                                                                  
$ hackrf_transfer -r /dev/null
```

Also, for double checking timing I've been running it with much shorter dwell times, so that any off-by-one would be clearly visible on the 'scope:
```
$ hackrf_operacake -o 0 -m time -t A1:10 -t A2:10 -t A3:10 -t A4:10
$ hackrf_transfer -r /dev/null -s 1e6
```

and the result looks good:
sctimer-timing png

### DIFF
--- a/host/libhackrf/src/hackrf.c
+++ b/host/libhackrf/src/hackrf.c
@@ -91,6 +91,7 @@ typedef enum {
 	HACKRF_VENDOR_REQUEST_UI_ENABLE = 37,
 	HACKRF_VENDOR_REQUEST_OPERACAKE_SET_MODE = 38,
 	HACKRF_VENDOR_REQUEST_OPERACAKE_GET_MODE = 39,
+	HACKRF_VENDOR_REQUEST_OPERACAKE_SET_DWELL_TIMES = 40,
 } hackrf_vendor_request;
 
 #define USB_CONFIG_STANDARD 0x1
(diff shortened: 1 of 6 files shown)


host/libhackrf/src/hackrf.c
```diff
@@ -2202,6 +2290,42 @@ int ADDCALL hackrf_set_operacake_ranges(hackrf_device* device, uint8_t* ranges,
 	}
 }
 
+static uint8_t dwell_data[5*16];
```
mossmann, line 2293: maybe use HACKRF_OPERACAKE_MAX_DWELL_TIMES here for clarity

mossmann: also maybe sizeof(hackrf_operacake_dwell_time)?

firmware/common/operacake.c
```diff
@@ -166,10 +229,14 @@ uint8_t operacake_set_ports(uint8_t address, uint8_t PA, uint8_t PB) {
...
 	pb = port_to_pins(PB);
-		
+
 	reg = (OPERACAKE_GPIO_DISABLE | side
 					| pa | pb | OPERACAKE_EN_LEDS);
 	operacake_write_reg(oc_bus, address, OPERACAKE_REG_OUTPUT, reg);
+
+	// Keep track of manual settings for when we switch in/out of time/frequency modes.
+	operacake_boards[address].PA = PA;
+	operacake_boards[address].PB = PB;
```
mossmann, line 239: Do we want operacake_set_ports() to behave differently if it is called while in time or frequency mode? It gets called by usb_vendor_request_operacake_set_ports() which maybe should stall if we're in one of those modes? Or maybe it should force the mode to manual?

mossmann (changes requested):
Excellent work! Thank you!

I made a few individual comments below.

host/libhackrf/src/hackrf.c
miek: I was reluctant to use sizeof(hackrf_operacake_dwell_time) there because I think it may differ from the size of the control transfer, depending on the packing choices of different compilers/hosts.
I could add a local define there just to make it clearer?

firmware/common/operacake.c
miek: I'm not sure about making a special case for set_ports / manual mode, I think it would be best to keep it consistent with the other modes, and I don't think we should automatically switch mode when setting dwell times or frequency ranges. I think it's reasonable to treat set_ports as a setting that only applies in manual mode, as long as we document that well.

PR #954, opened 2021-09-15
update mentions page

straithe:
Remove dead links from projects and mentions page.


epenelope (approved):
I checked all the links, looks good @straithe!

PR #971, opened 2021-10-15
cmake/modules/FindFFTW.cmake: fix build without fftw3

ffontaine:
Build on Linux fails if libfftw3 is not available since commit https://github.com/greatscottgadgets/hackrf/commit/a8c1fc92e9fa10843cb4aafd68c52f9903ebd05d which replaced `pkg_check_modules(FFTW REQUIRED fftw3f)` by `find_package(FFTW REQUIRED)`

Fix this build failure by updating `FindFFTW.cmake` to check for `fftw3f`

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


mossmann:
Thank you for this fix!

Issue #180, opened 2015-06-09
Firmware should stop TX/RX when communication is lost


kpreid:
Feature request:

If the host process which is controlling the HackRF is abruptly terminated, the HackRF remains in TX or RX mode. It would be nice if the firmware would instead recognize this condition and shut off the transmitter/receiver.

I'm not familiar enough with USB to know whether this is a detectable event (analogous to a lost TCP connection) or would require timeout logic (e.g. not receiving any TX data packets or other keepalive).

┆Issue is synchronized with this [Basecamp todo](https://3.basecamp.com/5295633/buckets/26082705/todos/4597360884) by [Unito](https://www.unito.io)

KC5CQW:
User adjustable "watchdog" timer is especially for TX is needed for compliance with various communications authorities. Not to mention, equipment/DUT safety.

rroff:
Hey,

Have you seen that latest movie? I don't know what all the fuss is about. Before you watch it read a review here <http://tusindutro.peterbillingsley.org/aecpdyn>

See you soon, rroff@roff.us

dominicgs:
For TX I plan to keep track of the freshness of the buffers to be transmitted. There's no sense in transmitting buffers that have already been transmitted (not in our current usage model), so I'll reset to idle if that happens.

I'm fairly sure it could work that way for RX, but I don't know if total failure is the answer for a slower than expected USB connection.  I think it's a fair thing to do, but I worry that it will break some things in subtle ways.

Issue #86, opened 2013-06-26
timed operations


mossmann:
Implement timed TX and RX operations: The host requests an operation to start and/or stop at a particular time in the future.

See also: https://github.com/mossmann/hackrf/issues/85

PR #982, opened 2021-11-03
Firmware sample buffer management overhaul, including safe handling of TX underruns

martinling:
This PR started out as an effort to fix bug #180, but has evolved into a more general overhaul of how the firmware's sample buffer is managed.

Previously, the M0 core would continuously and unconditionally transfer data between SGPIO and the buffer, whilst the M4 core tried to keep up with transferring that data between the buffer and the host to keep up with the M0. In the event of a TX underrun (not enough data from the host), old buffer contents would be retransmitted. In the event of an RX overrun (host not retrieving data quickly enough), old data would be overwritten. There was no way for the host to detect that either of these problems had occured. If the host crashed or USB communication was lost, TX could continue with existing buffer contents indefinitely.

This PR replaces the simple SGPIO loop on the M0 with a more complex one, which keeps track of the available bytes in the buffer and stops advancing if a TX underrun or RX overrun would occur. In the case of an RX overrun, new data is discarded instead of overwriting existing data. In the case of a TX underrun, zeroes are sent to the DAC instead of repeating previous buffer contents. This ensures that **nothing further is transmitted if the USB stream ceases**.

In addition, the M0 code now maintains statistics on the buffer usage, including the number of underruns or overruns, and the length of the longest in bytes. These statistics make it possible to diagnose USB throughput problems. The host can retrieve the buffer statistics via a USB vendor request which retrieves the M0 state, and an associated libhackrf function:  `hackrf_get_m0_state()`.

Two new features have been added to the host tools to make use of the data provided by the M0:

* A new `hackrf_debug -S` option to retrieve and display the current M0 state. This can be used to check buffer statistics after running existing programs which are not themselves aware of the new API. The stats displayed will reflect the last TX or RX operation, e.g:

```
$ hackrf_debug -S
M0 state:
Mode: 0 (IDLE)
M0 count: 21266400 bytes
M4 count: 21233664 bytes
Number of shortfalls: 1
Longest shortfall: 576 bytes
Shortfall limit: 2048 bytes
Mode change threshold: 0 bytes
Next mode: 0 (IDLE)
```

* A new `hackrf_transfer -B` option which adds buffer statistics to the information printed by `hackrf_transfer`, e.g:

```
$ hackrf_transfer -r /dev/null -s 20000000 -B
call hackrf_set_sample_rate(20000000 Hz/20.000 MHz)
call hackrf_set_freq(900000000 Hz/900.000 MHz)
Stop with Ctrl-C
39.8 MiB / 1.000 sec = 39.8 MiB/second, amplitude -19.8 dBfs, 16224 bytes free in buffer, 0 overruns, longest 0 bytes
40.1 MiB / 1.000 sec = 40.1 MiB/second, amplitude -19.8 dBfs, 19808 bytes free in buffer, 0 overruns, longest 0 bytes
39.8 MiB / 1.000 sec = 39.8 MiB/second, amplitude -19.8 dBfs, 23264 bytes free in buffer, 1 overruns, longest 544 bytes
40.1 MiB / 1.000 sec = 40.1 MiB/second, amplitude -19.8 dBfs, 27392 bytes free in buffer, 1 overruns, longest 544 bytes
40.1 MiB / 1.000 sec = 40.1 MiB/second, amplitude -19.9 dBfs, 15712 bytes free in buffer, 1 overruns, longest 544 bytes
39.8 MiB / 1.001 sec = 39.8 MiB/second, amplitude -19.8 dBfs, 20736 bytes free in buffer, 1 overruns, longest 544 bytes
40.1 MiB / 1.000 sec = 40.1 MiB/second, amplitude -19.8 dBfs, 27424 bytes free in buffer, 2 overruns, longest 544 bytes
40.1 MiB / 1.000 sec = 40.1 MiB/second, amplitude -19.9 dBfs, 31584 bytes free in buffer, 2 overruns, longest 544 bytes
39.8 MiB / 1.001 sec = 39.8 MiB/second, amplitude -16.8 dBfs, 30688 bytes free in buffer, 2 overruns, longest 544 bytes
40.1 MiB / 1.000 sec = 40.1 MiB/second, amplitude -19.8 dBfs, 22432 bytes free in buffer, 2 overruns, longest 544 bytes
^CCaught signal 2
22.8 MiB / 0.569 sec = 40.1 MiB/second, amplitude -19.8 dBfs, 23424 bytes free in buffer, 3 overruns, longest 608 bytes

Exiting...
Total time: 10.57293 s
hackrf_stop_rx() done
Transfer statistics:
422955712 bytes transferred by M0
422936576 bytes transferred by M4
5 overruns, longest 864 bytes
hackrf_close() done
hackrf_exit() done
fclose() done
exit
```

The host can also set limits on the maximum length of TX underruns and RX overruns, in bytes, via two additional libhackrf functions, `hackrf_set_tx_underrun_limit` and `hackrf_set_rx_overrun_limit`. These limits can also be set by new `-T` and `-R` options in `hackrf_debug`. The default TX underrun & RX overrun limits are set to zero, which disables the limits. Any other value is interpreted as a limit in bytes. After a contiguous gap of this many bytes in the TX sample stream, or a contiguous loss of this many bytes from the RX stream, the M0 will stop transferring samples.

Setting these limits makes it possible to ensure that TX or RX will stop completely if the host fails to sustain a sufficient flow of sample data over USB. If a configured limit is exceeded, the M0 code will cease transferring samples, which will block the M4 from making any further USB transfers to or from the buffer, thereby stalling the host side.

This PR also adds the necessary M0-side support for timed operations (#86), and reimplements sweep mode using this feature. Doing this turned out to be necessary in order to make sweep mode play nicely with the new buffer management. The new version also improves sweep mode reliability by making it less sensitive to host throughput. Sweep timing is now guaranteed to be consistent provided that the host can keep up with an average 6.67Msps transfer rate with no more than 10ms between 16KBtransfers. If the host fails to keep up, sweeping will now pause for the host to catch up, rather than data being lost or corrupted.


gozu42:
i did some testing with this on a raspi4.
tests involved random-receive with hackrf_transfer at different sample rates, with one hackrf connected directly, the other through a hub.
no relevant difference observed related to the hub-or-not part of the setup.

one minor note ...
```
$ hackrf_info
hackrf_info version: git-adddca3
libhackrf version: git-adddca3 (0.7)
Found HackRF
Index: 0
Serial number: 0000000000000000088869dc382f661b
Firmware Version: git-adddca3 (API:1.06)

$ time hackrf_transfer -B -r /dev/null -f 97300000 -s 10000000 -p 0 -a 0 -l 0 -g 0 -n 50000000 -d c382f661b
call hackrf_set_sample_rate(10000000 Hz/10.000 MHz)
call hackrf_set_freq(97300000 Hz/97.300 MHz)
call hackrf_set_amp_enable(0)
call hackrf_set_antenna_enable(0)
samples_to_xfer 50000000/50Mio
Stop with Ctrl-C
19.9 MiB / 1.001 sec = 19.9 MiB/second, amplitude -17.1 dBfs, 13792 bytes free in buffer, 0 overruns, longest 0 bytes
19.9 MiB / 1.001 sec = 19.9 MiB/second, amplitude -17.1 dBfs, 19744 bytes free in buffer, 0 overruns, longest 0 bytes
20.2 MiB / 1.005 sec = 20.1 MiB/second, amplitude -17.1 dBfs, 17408 bytes free in buffer, 0 overruns, longest 0 bytes
20.2 MiB / 1.002 sec = 20.2 MiB/second, amplitude -17.1 dBfs, 15200 bytes free in buffer, 0 overruns, longest 0 bytes
19.9 MiB / 1.004 sec = 19.9 MiB/second, amplitude -14.1 dBfs, 16096 bytes free in buffer, 0 overruns, longest 0 bytes

Exiting... hackrf_is_streaming() result: streaming terminated (-1004)
Total time: 5.01317 s
hackrf_stop_rx() done
Transfer statistics:
100368352 bytes read from SGPIO by M0, 100335616 sent to host by M4
Buffer margins: max 32768, min 0 bytes free, 1 overruns, longest 6944 bytes
hackrf_close() done
hackrf_exit() done
fclose() done
exit

real    0m5.106s
user    0m1.234s
sys     0m0.206s
$ time hackrf_transfer -B -r - -f 97300000 -s 10000000 -p 0 -a 0 -l 0 -g 0 -n 50000000 -d c382f661b | buffer | wc
call hackrf_set_sample_rate(10000000 Hz/10.000 MHz)
call hackrf_set_freq(97300000 Hz/97.300 MHz)
call hackrf_set_amp_enable(0)
call hackrf_set_antenna_enable(0)
samples_to_xfer 50000000/50Mio
Stop with Ctrl-C
 9.7 MiB / 1.001 sec =  9.7 MiB/second, amplitude -17.1 dBfs, 17696 bytes free in buffer, 29 overruns, longest 342944 bytes

Exiting... hackrf_is_streaming() result: streaming terminated (-1004)
Total time: 11.49984 s
hackrf_stop_rx() done
Transfer statistics:
100958176 bytes read from SGPIO by M0, 100925440 sent to host by M4
Buffer margins: max 32768, min 0 bytes free, 375 overruns, longest 354944 bytes
hackrf_close() done
hackrf_exit() done
exit  
      0       0 100000000

real    0m11.688s
user    0m12.149s
sys     0m0.821s

```

note the difference in stats output between the "straight to devnull" and "to stdout" runs.
for the "to stdout" run, the stats output that shows up _during_ the run is actualy truncated after dBfs, and the buffer stats show only up at the end.
and the throughput/runtime is ... very much different too?
while this is running, the wc is cpu-bound.
if i change the pipe to "| buffer > /dev/null", it reverts to the "straight to devnull" performance and output.
if i crank up the buffer sizes (-s 131072 -m 268435456) it goes ok too.
so somehow the loop-progress (and stats) output get impacted by the blocking/limited output.

with current HEAD:
```
$ hackrf_info
hackrf_info version: git-22267f3
libhackrf version: git-22267f3 (0.6)
Found HackRF
Index: 0
Serial number: 0000000000000000088869dc382f661b
Firmware Version: git-22267f3 (API:1.05)

$ time hackrf_transfer -r - -f 97300000 -s 10000000 -p 0 -a 0 -l 0 -g 0 -n 50000000 -d c382f661b | buffer | wc
call hackrf_set_sample_rate(10000000 Hz/10.000 MHz)
call hackrf_set_freq(97300000 Hz/97.300 MHz)
call hackrf_set_amp_enable(0)
call hackrf_set_antenna_enable(0)
samples_to_xfer 50000000/50Mio
Stop with Ctrl-C
10.0 MiB / 1.001 sec = 10.0 MiB/second, amplitude -17.1 dBfs
 8.7 MiB / 1.000 sec =  8.6 MiB/second, amplitude -17.1 dBfs
 8.9 MiB / 1.001 sec =  8.9 MiB/second, amplitude -17.1 dBfs
 8.4 MiB / 1.001 sec =  8.4 MiB/second, amplitude -17.1 dBfs
 8.7 MiB / 1.001 sec =  8.6 MiB/second, amplitude -17.1 dBfs
 8.4 MiB / 1.000 sec =  8.4 MiB/second, amplitude -17.1 dBfs
 8.7 MiB / 1.000 sec =  8.6 MiB/second, amplitude -17.2 dBfs
 8.4 MiB / 1.000 sec =  8.4 MiB/second, amplitude -14.0 dBfs
 8.7 MiB / 1.000 sec =  8.6 MiB/second, amplitude -17.1 dBfs
 8.7 MiB / 1.000 sec =  8.6 MiB/second, amplitude -17.1 dBfs
 8.7 MiB / 1.000 sec =  8.6 MiB/second, amplitude -17.1 dBfs
 5.0 MiB / 1.000 sec =  5.0 MiB/second, amplitude -18.0 dBfs

Exiting... hackrf_is_streaming() result: streaming terminated (-1004)
Total time: 12.00447 s
hackrf_stop_rx() done
hackrf_close() done
hackrf_exit() done
exit  
      0       0 100000000

real    0m12.106s
user    0m12.142s
sys     0m0.802s
```

note the data output runs into the same issue/constraint (lower-than-expected MiB throughput), but the stats happen just fine.

martinling:
@gozu42, I think the differences you're running into there must result from the fact that with the `-B` option to report stats, `hackrf_transfer` now has to make some `libhackrf` calls from the main thread during the transfer, which in turn make USB vendor requests to get the stats.

Specifically, if `-B` is given, it calls `hackrf_get_buffer_stats` _after_ printing the line up to `dBfs`, and when the stats are retrieved they're appended to the line:

```
fprintf(stderr, "%4.1f MiB / %5.3f sec = %4.1f MiB/second, amplitude %3.1f dBfs",
        (byte_count_now / 1e6f),
        time_difference,
        (rate / 1e6f),
        dB_full_scale_ratio);
if (display_stats) { 
        hackrf_buffer_stats stats;
        result = hackrf_get_buffer_stats(device, &stats);
        if (result != HACKRF_SUCCESS)
                fprintf(stderr, "\nhackrf_get_buffer_stats() failed: %s (%d)\n", hackrf_error_name(result), result);
        else
                fprintf(stderr, ", %d bytes %s in buffer, %d %s, longest %d bytes\n",
                        (transmit || signalsource) ? stats.m4_count - stats.m0_count : stats.m0_count - stats.m4_count,
                        (transmit || signalsource) ? "filled" : "free",
                        stats.num_shortfalls,
                        (transmit || signalsource) ? "underruns" : "overruns",
                        stats.longest_shortfall);
} else { 
        fprintf(stderr, "\n");
}
```

So it seems that in the case where you're piping the output through `buffer | wc`, that `hackrf_get_buffer_stats` call blocks, and doesn't return until the whole transfer has completed - which means no further iterations through the main loop are made, and no further per-second throughput figures are printed.

I'm not sure why that call should block - it's shouldn't be sending anything to `stdout`, so it shouldn't care if the pipe is blocked. It would be interesting to attach gdb and see where it's stuck.

I think you should find that if you _don't_ pass the `-B` flag, the behaviour should be the same as with master.

gozu42:
> I think you should find that if you _don't_ pass the `-B` flag, the behaviour should be the same as with master.

i can confirm this is the case.

martinling:
@gozu42 I can't reproduce the problem exactly here on x86 - the same command line (minus the specific HackRF serial number) works correctly for me. But when I step it up to 20Msps, I do see a brief delay between the `dBfs` and the buffer stats being printed. So I think the issue may be that the system is struggling to get the vendor request processed with all the bulk transfer traffic flying.

gozu42:
> So I think the issue may be that the system is struggling to get the vendor request processed with all the bulk transfer traffic flying.

when you get to the "looking at the usb traffic with luna" part, i would be very interested in your setup/code/tooling for that so i can do the "same" measurement/test/logging here.

martinling:
I had previously failed to account for the SGPIO-to-AHB bridge latencies when reading and clearing the interrupt status, and taking these into account put the theoretical worst case cycle timings just over the limit - 167 cycles in the RX path, against a limit of 163.

This did not cause any problems in practice - the code passed all tests with the SGPIO debug image for the CPLD.

However, to be safe, I have shaved off another 7 cycles through various tricks, and updated the comments and cycle count annotations to show that the code is now within the worst-case limits. The 5 commits just pushed add these changes.

martinling:
Unless any new problems are spotted, I don't plan to make any further changes to this PR as it stands.

As @gozu42 has observed, there can be delays when trying to retrieve the buffer stats during a transfer. This is not a regression however - there is no change to the behaviour of `hackrf_transfer` without the new `-B` flag. It seems to simply be the result of the system struggling to get a USB vendor request through when already saturated with bulk transfers (and still not keeping up with the requested sample rate). It's not a bug, but just a symptom of the system failing to keep up with what it's been asked to do.

The best way I can see to avoid that particular problem would be to include the buffer statistics along as metadata in the bulk transfer stream, rather than needing a separate vendor request. Adding metadata to the stream is on my roadmap, but out of the scope of this PR.

gozu42:
looks good, it seems more stable at barely-viable sample rates now.

i have one request with this still open: trying to avoid counting end-of-streaming as a massive shortfall.
we talked about this before, and i understand _why_ this happens, but it still means the "longest shortfall" reported by "hackrf_debug -S" is ... fairly meaningless currently.
did you look into whether doing the shutdown in a different order/manner would be enough to avoid it?
or some kind of "in shutdown" flag that blocks updating stats?

martinling:
> i have one request with this still open: trying to avoid counting end-of-streaming as a massive shortfall. we talked about this before, and i understand _why_ this happens, but it still means the "longest shortfall" reported by "hackrf_debug -S" is ... fairly meaningless currently. did you look into whether doing the shutdown in a different order/manner would be enough to avoid it? or some kind of "in shutdown" flag that blocks updating stats?

I've looked at this, but I don't see a good way to fix it without larger changes.

Right now, `hackrf_stop_rx()` looks like this - it stops the transfers first, and then calls `hackrf_stop_rx_cmd()`, which sends the request to drop out of RX mode:
```c
int ADDCALL hackrf_stop_rx(hackrf_device* device)
{
        int result;

        device->streaming = false;
        result = cancel_transfers(device);
        if (result != HACKRF_SUCCESS)
        {
                return result;
        }

        return hackrf_stop_rx_cmd(device);
}
```
If we put this in the other order - i.e. request the stop first, then cancel the bulk transfers - then on a slow host we risk running into the problem you saw earlier. When the bus was saturated with bulk transfers, `hackrf_get_buffer_stats()` got blocked, because it couldn't get a USB vendor request through. If we try to make a stop request under the same conditions, it would block in the same way, and then `hackrf_stop_rx()` would just hang without ever getting to `cancel_transfers()`.

So we do need to stop the bulk transfers first, to make sure that the stop request actually gets through.

And from the device side, that means that the sample stream can dry up before the stop request actually gets seen. Which is indistinguishable from a shortfall.

I think the only way to solve this reliably is to piggyback stop requests on the bulk transfer traffic. For TX, that could be done in metadata within the buffer data ("here's the next N samples, and stop after sending these"). For RX it's a bit more complicated, and I'm not sure what the best way to do it is yet.

martinling:
> and currently the stats are very misleading for hackrf_sweep, probably because it doesnt distinguish between the 2/3 of blocks thrown away and the 1/3 blocks that actualy get sent over to host.

I've had a look at this, but I'm not sure what's misleading about it. Do you mean the byte counts?

The counts in the stats after a `hackrf_sweep` call correctly reflect the number of bytes written to the buffer by the M0, and the number read from the buffer by the M4. These are the meanings documented for those counts [here](https://github.com/greatscottgadgets/hackrf/pull/982/files#diff-e4d18e38ea1b9068f7e9fac04eb7bb2df084d1425fec3992025b893ceab7c529R172).

For sweep mode, some of that data is discarded by the M4 without being sent to the host, but that doesn't make the count incorrect, since it's not a count of bytes sent to the host.

In `hackrf_transfer`, I do report the M4 count as "bytes sent to host", which is correct within that use case - but I don't report the count in `hackrf_sweep`, and I don't use that description for it in `hackrf_debug --stats` - it's just labelled as `M4 count`.

It's not straightforward to change the meaning of these stats, because they're used directly by the M0 code, and retrieved directly from memory by the USB buffer stats request in order to get the latest figures.

martinling:
For the spurious shortfalls at shutdown, I'm going to try to detect this in the M0 code by spotting when the host changes the mode to `IDLE` during a shortfall, and disregarding that shortfall in the stats if this happens.

gozu42:
i mainly mean this part (after a -N 10 sweep):

> Number of shortfalls: 7517
> Longest shortfall: 19008 bytes

the "longest shortfall" is a variant of the "shutdown" problem. 
but (when using these stats as a way to diagnose host/usb performance issues) i would expect the "number of shortfalls" to be 0 (or 1, considering the shutdown problem) ... because there wasnt a single block lost that was meant to be transferred.

martinling:
@gozu42 I believe those are actually genuine shortfalls encountered during the sweep. I can run with `-N 10` and get no shortfalls reported.

When the sweep code discards a block, it advances `m4_count` so that the M0 knows it has consumed that block - see [this line](https://github.com/greatscottgadgets/hackrf/pull/982/files#diff-399dab3793596f7bb1213c1d6c57f479b83acdeadfcbbbb16dc4da6036f639b7R143).

gozu42:
> @gozu42 I believe those are actually genuine shortfalls encountered during the sweep. I can run with `-N 10` and get no shortfalls reported.

what does "genuine shortfall" mean in this case though? 
the setup i am testing on can handle 16Msps with hackrf_transfer with no significant shortfalls.
hackrf_sweep does 20Msps, which is well above that, but also only transfers 1/3 of the samples to host.
since we have only two buffers, lets say effectively 10Msps (instead of 6.66Msps) because we need to be done with the buffer before it gets reused.
it should be very much able to handle that? 
is there some delay/penalty for "stop and go" instead of "continous streaming"?

martinling:
> what does "genuine shortfall" mean in this case though?

It means the M0 had read some samples from SGPIO and had to discard them, because there was no space in the buffer.

In the sweep case there are two ways for the samples to leave the buffer - either the M4 discards a block, or the host finishes transferring one. The discard should be close to instantaneous, so most likely the shortfalls are due to delays in the host retrieving the data.

Note that even though the overall USB data rate should be lower for sweep mode, the latency with which the host retrieves each block that is transferred is also critical.

gozu42:
one possible difference between hackrf_transfer and hackrf_sweep is that _sweep does a LOT more work (fft) on the rx_callback.
@mossmann suggested i try with different binwidth (which should change cpu need of the fft part).
the results dont show any useful pattern. 10 runs for different fft sizes from 10k to 500k:
```
10k: 7694 6928 8550 9204 6565 9022 7954 7839 7604 8578
50k: 9372 7471 9343 9352 8416 6960 9344 7624 8474 9194
100k: 5458 7631 9606 6542 6759 8742 3793 8511 8695 9514
500k: 8975 6668 7570 8407 8549 9337 9179 8540 8438 8373
```

so LGTM, and i will toy around with offloading the fft-work on sweep from the rx-callback/thread and file a new issue if/when needed.

martinling:
Whilst trying to solve the "shortfalls at shutdown" problem, I have discovered a new one. :-(

When the M4 shuts down SGPIO streaming, the M0 is left spinning on the SGPIO interrupt flag. So the M0 doesn't actually get an opportunity to detect that it's been switched back to `IDLE`. It doesn't leave that loop until the M4 has already switched it back to `RX` or `TX_START` and re-enabled SGPIO streaming.

So as it stands, there's nowhere I can put any code that rolls back the shortfall stats if a shutdown is detected. And the code between `idle` and `loop` doesn't actually get executed, except at startup.

A solution is to have the M0 spin on both the `mode` value and the SGPIO interrupt flag, but that adds several cycles to the loop. I'm already having problems fitting in the extra bookkeeping for rollback, especially in the TX underrun path, which becomes the critical one with the rollback features.

I might be able to make it work, but unless there's some other breakthrough, I'm not going to have any cycles left to implement the chunking needed for larger buffer sizes.

It would be really good to work out why we can't just poll the NVIC or use a `wfi`. It would save a couple dozen cycles.

martinling:
I found a solution - after setting the M0 to idle, the M4 can just set the SGPIO interrupt flag itself, which lets the M0 code run again and gives it the chance to notice that it's been set to idle.

With that change, I've now implemented rollback of the final shortfall, in the case where the M0 is reset to idle whilst that shortfall is still ongoing. This seems to eliminate spurious shortfalls at shutdown - @gozu42, you may want to test this.

One shortcoming of the approach is that when this happens, the minimum buffer margin will still show up as zero, because we don't back up the stats for rollback until there's actually a shortfall. I don't see an easy way to avoid that - I don't see how you could identify the last "normal" level.

gozu42:
i can confirm this seems to fix the "shutdown shortfall" issue for the hackrf_transfer case.

but i am still/again confused by the shortfalls reported after hackrf_sweep.
i tried various things by now:

if i disable all processing of data in the hackrf_sweep rx_callback (no copying, no fft, no printing), leaving just the bare minimum housekeeping, it doesnt seem to have any impact on the reported shortfalls at all.

if i disable the m4_count callback in sweep_mode (and increase m4_count right after scheduling the transfer), i get 0 shortfalls, so it is really tied to USB timing.

if i disable the dis/en for the usb-irq around set_freq in sweep_mode (which will happen after _every_ transfer being scheduled), it doesnt seem to change anything.

if i disable transfer of buffer 1 in rx_mode (which means it transfers 1/2 buffers, compared to 1/3 in sweep_mode), i get stable results with 20Msps on hackrf_transfer (and the expected half bandwidth). 
so this really seems to be specific to sweep_mode in some way, and not a bandwidth/latency limitation as such.

also just realized something about the test with the different binwidth ... a lot of these runs had more shortfalls than blocks were transferred. one 0-6Ghz sweep in interleaved mode is 600 blocks. at -N10, it should be 6000 blocks (plus shutdown-overrun). yet a lot of these reported shortfalls of 9000+?

martinling:
@gozu42 The thing is that I haven't been able to reproduce those shortfalls on my x86 system - I get zero shortfalls for `hackrf_sweep -N 10`, both with and without the shutdown changes. If it was a bug in the buffer management rather than a throughput/latency issue, I'd expect see it here too.

I wonder if the extra shortfalls (~9000 vs 6000) are due to the HackRF having continued to run further sweeps (up to a total of 15 or so), whilst the host is still trying to read out the first 10? Timing the command could confirm that - with the default settings it should be about 0.75s per sweep:
```
$ time hackrf_sweep -r /dev/null -N10
call hackrf_sample_rate_set(20.000 MHz)
call hackrf_baseband_filter_bandwidth_set(15.000 MHz)
Sweeping from 0 MHz to 6000 MHz
Stop with Ctrl-C
1 total sweeps completed, 1.00 sweeps/second
2 total sweeps completed, 1.00 sweeps/second
4 total sweeps completed, 1.33 sweeps/second
5 total sweeps completed, 1.25 sweeps/second
6 total sweeps completed, 1.20 sweeps/second
8 total sweeps completed, 1.33 sweeps/second
9 total sweeps completed, 1.28 sweeps/second

Exiting...
Total sweeps: 10 in 7.41369 seconds (1.28 sweeps/second)
hackrf_close() done
hackrf_exit() done
fclose() done
exit

real	0m7.454s
user	0m0.168s
sys	0m0.086s
```

Is this still on RPi4? I don't have one here, but could try to get hold of one to reproduce.

martinling:
Another thought is that it might be worth stopping the M0 while retuning. As it stands, the M0 is still going to be reading from SGPIO and trying to put things in the buffer whilst this happens:

```c
nvic_disable_irq(NVIC_USB0_IRQ);
set_freq(sweep_freq + offset);
nvic_enable_irq(NVIC_USB0_IRQ);
```
That wouldn't have mattered so much before my changes because the M0 would just continue regardless, and the affected data would get thrown away afterwards. But now the M0 gets to complain.

I don't recall how fast a retune is, but it involves some I2C, and the buffer is less than a millisecond's worth of runway.

gozu42:
this is still on a raspi 4, yes.

> That wouldn't have mattered so much before my changes because the M0 would just continue regardless, and the affected data would get thrown away afterwards. But now the M0 gets to complain.

this seems to be the issue.
after applying your suggested "stop m0 for retune" change from discord, it looks a lot better.

100 "-N 10" runs with current bug-180:
```
8x <3000 shortfalls, 23x <6000, 69x <9000
7x 1.28 sweep/s, 23x 1.12 sweep/s, 70x 1.00 sweep/s ...
```

51 runs with setting m0 to idle for retune:
```
46x 0 shortfalls, 5x 1 shortfall.
50x 1.12 sweep/s, 1x 1.24 sweep/s
```

100x with origin/HEAD:
```
(obviously no reported shortfalls)
100x 1.28 sweep/s
```
 
> I don't recall how fast a retune is, but it involves some I2C, and the buffer is less than a millisecond's worth of runway.

it is so slow that the interesting SPI variant used to configure the mixer would end up in an undefined state if a hackrf_stop_rx arrived while do_sweep was retuning. thats the reason for disabling the USB IRQ there.

now the "many more shortfalls than trasferred blocks" make sense too? 
because if the m0 "fully laps" the m4, like being all the way through a buffer before do_sweep sees it while being distracted with retune ... the throwaway buffers can end up being counted as shortfalls too, and it explains why the overall sweep rate went down. (because m4 is waiting for m0 to get ahead of it again)

martinling:
I've now added the fix discussed for the latency sensitivity seen with hackrf_sweep, and merged the latest changes from master which had caused a conflict for the library version changes in CMakeLists.txt. It's looking like this may be good to go now.

gozu42:
did 100 "-N 10" sweep runs with this version.

92 runs didnt have any shortfalls, with min buffer margins between 704 and 0 bytes. (living on the edge)
7 runs had one shortfall of 32 bytes in all cases. (the other side of that edge)
1 run fully reverted to the "bad" situation, with 5702(*) reported shortfalls, ups.

that last part makes me suspect that perhaps tuning isnt the problem after all, and the change is just buying my barely-stable setup the few bytes of additional margin it needs to keep up.
but tuning hackrf_sweep is really a mission for another issue/pr, and this one gives us a _very_ powerful tool / metrics for that!

so, looks good to me, and :+1: for merge. 

(*) its still suspicious to me that its either "all good" or "all hell", with no observed "middle ground" cases. either it gets very confused about buffer management (m4 getting lapped by m0) or once things went south it straight up doesnt recover from that (for lack of margin). but again, mission for another day.

martinling:
The conclusion of recent discussions was that we want to make sure this PR doesn't change the timing behaviour of sweep mode. 

The quick fix to stop the M0 during retuning doesn't achieve that, and it turns out to be surprisingly complicated to come up with a better fix. So I'm doing some more work towards that end.

The roadmap I'm working to at the moment is:

1. Restructure the assembly code using separate per-mode loops, with shared code in macros. This will make it easier to add new modes, without extra cycles for tests & branches, and to change behaviour that's common to multiple modes.

2. Add a `WAIT` mode in which the M0 will advance `m0_count` but not write to the buffer or access SGPIO.

3. Add a setting for the M0 to switch modes upon reaching a target `m0_count` value. This also lays some groundwork for implementing #86.

4. Use the new features to reimplement sweep mode as follows:

    0. M4 initially puts the M0 into `RX` mode, with an `m0_count` threshold of 16K and a next mode of `WAIT`.
    1. M4 spins until the M0 switches to `WAIT` mode.
    2. M0 captures one 16K block of samples, and switches to `WAIT` mode.
    3. M4 sees the mode change, advances the `m0_count` target by 32K, and sets next mode to `RX`.
    4. M4 adds the sweep metadata at the start of the block and schedules a bulk transfer for the block.
    5. M4 retunes - this takes about 760us worst-case, so should be complete before the M0 goes back to RX.
    6. M4 spins until the M0 mode changes to `RX`, then advances the `m0_count` limit by 16K and sets the next mode to `WAIT`.
    8. Process repeats from step i.

The sweep data will therefore alternate between the two 16K blocks of the USB buffer area, giving plenty of leeway for delays in the host completing the bulk transfers.

martinling:
Just as a heads up for anyone who's been using this - I'm going to be rebasing and cleaning up the history on this branch soon to make the changes easier to follow and review. So there may be some churn and force pushes for a while; I'll let people know here when it's ready for review again.

martinling:
Scheduled mode changes are now implemented between RX & WAIT modes, and I've reimplemented sweep mode on top of this feature, exactly as proposed previously. The code is right on the cycle count limit, but it seems to all be working.

I have rebased the `bug-180` branch on `master`, so it no longer includes unrelated changes.

I have not yet cleaned up the history to make it easier to review step-by-step. However, the overall diff against master does reflect the intended overall changes. Testing of the current head commit `7924c10`, especially using `hackrf_sweep`, would be welcome.

Meanwhile, I am going to attempt to clean up the commit history to make it easier to review, so the `bug-180` branch may be force-pushed again soon.

martinling:
I've done some further work to make the assembly easier to read. In the process, I turned up some bugs and fixed them.

straithe:
Love it! You are doing great.

gozu42:
bunch of testruns with a clean build up to 8372ca9 ... still on the raspi4, so "marginal host performance". 

runs with 50k, 100k, 500k, 1M binwidth look very good/consistent. zero shortfalls, very consistent 1.28sweep/sec.

at 10k binwidth i am seeing shortfalls, but these seem to be related to the host actually blocking. ~300 shortfalls with sizes up to 500k bytes. 
if i redirect the output straight to devnull, the 10k binwidth also returns mostly zero shortfalls. one run out of 20 was still reporting 2 shortfalls, with a much more reasonable 11k size.
if i put heavy IO load on the rpi4 during a 10k run, i am able to get more shortfalls even with writing to devnull, so it seems i am simply maxing out both the cpu _and_ the usb io at that point. 

the reported sweep performance of 1.28s/s is higher than what we had two weeks ago with the "stop m0 during retune" (as expected). going back to old logs (march-to-may timeframe), the reported sweep rates back then were 1.35s/s, but this feels like a change in reporting or even "it is consistent now instead of silently causing weird underruns/corruption every now and then". 

i reviewed the firmware sweep code, looks solid. the m0 code is, tbh, way over my head. 

overall: LGTM, very solid improvement over the last two weeks!

martinling:
I'm considering getting rid of the max/min buffer margin stats.

The values are usually not that interesting - if everything else is done well, the maximum margin should be the buffer size. That's the case at the start of RX, and will be the case in TX if the buffer is fully pre-filled, which I think it will be after we get some version of #662 merged. And the minimum will often be zero - either because the M0 was given a fixed amount to transmit/receive (which I'd like to arrange for `hackrf_transfer -n` to do), or because the user hit Ctrl-C and the margin ran out during shutdown.

It costs 10 worst-case cycles to update the max/min margin figures, which is currently the only work in the RX/TX loops that is really overhead just for the sake of statistics. The M0 and M4 counts are integral to the buffer management. The shortfall stats (number of shortfalls, and longest shortfall) are only updated in the shortfall paths, which aren't the critical ones for timing. 

I think we have better things to do with those 10 cycles. This PR is currently right at the limit of 163 cycles in RX. My proof-of-concept branch that increases the bulk buffer size to 64K, requires 7 extra cycles to arrange for the chunking necessary.

I think being able to get that bigger buffer is more useful than having extra stats.

Any thoughts?

martinling:
I've now added several pages of documentation to the M0 code, to explain what it does and how it works.

Also, by reordering the various routines, I've managed to set things up so that all the loops are close enough to each other to make conditional branches to each other. So the TX mode can now be used in timed operations, not just RX and WAIT as I had before to implement sleep mode with.

This should make the M0 code sufficient for implementing #86.

gozu42:
> I'm considering getting rid of the max/min buffer margin stats.
[...]
> I think being able to get that bigger buffer is more useful than having extra stats. 

:+1: on that, in particular since these stats are not all that useful to begin with.

for the sweep usecase, the max is always 32k-32, the min is either 16k (no shortfalls) or 0 (with shortfalls). 
all these numbers are fully expected/predictable given the "idle" mechanic, so no point burning precious cycles on them.

gozu42:
some testing with 498b370 on both the rpi4 and a x86 host with more oomph. works really well.

minor/cosmetic nitpick:
```
$ hackrf_debug -S
M0 state:
Mode: IDLE
M0 count: -1343849472 bytes
M4 count: -1343832064 bytes
Number of shortfalls: 1
Longest shortfall: 576 bytes
Shortfall limit: 0 bytes
Mode change at: 0 bytes
Next mode: IDLE
```

this is after a sweep with -N 100.
perhaps make the count numbers printed there unsigned. (unless there is a subtle meaning in them being negative)

gozu42:
ca 1.5 additional observations. oomphy x86 host, but with virtualized usb right now, so some shortfalls are not unexpected.

```
$ time hackrf_transfer -B -r /dev/null -f 97300000 -s 20000000 -p 0 -a 0 -l 0 -g 0 -n 1500000000
[...]
40.1 MiB / 1.002 sec = 40.0 MiB/second, amplitude -17.9 dBfs, 15360 bytes free in buffer, 6 overruns, longest 54592 bytes
33.8 MiB / 1.001 sec = 33.8 MiB/second, amplitude -17.9 dBfs, 32768 bytes free in buffer, 7 overruns, longest 5564992 bytes

Exiting... hackrf_is_streaming() result: streaming terminated (-1004)
Total time: 75.15133 s
hackrf_stop_rx() done
Transfer statistics:
-1293910016 bytes read from SGPIO by M0, -1293942784 sent to host by M4
[...]
$ hackrf_debug -S
M0 state:
Mode: IDLE
M0 count: -1293910016 bytes
M4 count: -1293942784 bytes
Number of shortfalls: 6
Longest shortfall: 54592 bytes
```

the last "inline" reported -B results have the "long shortfall during shutdown".
the "after shutdown" hackrf_debug stats have the "cleaned up" shortfall number.
this _might_ go away on its own with a "proper" fix for #1019 since it reeks at least in part of an unfortunate interaction between the rx callback, -n, and the 1sec-granularity main loop. 
unless you have any immediate ideas about this, i would call it an artifact out of scope for this PR.

and hackrf_transfer also reports the count-stats as signed.

both again cosmetic/nitpicky "can i have this icon in cornflower blue" remarks.
and another massive :+1: from over here for an awesome tool/feature that helps to understand what ones hackrf (and overall system) are doing.

martinling:
> perhaps make the count numbers printed there unsigned. (unless there is a subtle meaning in them being negative)

Good catch, thanks! I'll fix that.

> the last "inline" reported -B results have the "long shortfall during shutdown".
> the "after shutdown" hackrf_debug stats have the "cleaned up" shortfall number.
> this _might_ go away on its own with a "proper" fix for #1019 since it reeks at least in part of an unfortunate interaction between the rx callback, -n, and the 1sec-granularity main loop.

I think the "proper" fix for the shutdown shortfalls is probably for the host to actually tell the HackRF *when* to stop.

For `hackrf_sweep -N`, that would mean adding a new vendor request that sets a "number of sweeps desired" value, and the M4 would count down that number and exit sweep mode when done. If we also add a counter and a get request for the "current sweep number", then the host could do:

1. Ctrl-C received.
2. Host sends vendor request setting number of remaining sweeps to zero.
3. Host sends vendor request getting current sweep number.
4. Host continues receiving until it has all the data for the current sweep number.

The same approach could also be used in `hackrf_transfer`, but with sample counts instead of sweep counts.

martinling:
A lot of the general cleanup & optimisation work in this PR is now redone in #1022, which is intended to be merged first to lay some groundwork for the new features. That's ready for review now, and I'll be rebasing the rest of this work on it.

martinling:
I've just pushed a new version of this branch with a cleaned-up history that should be easier to review change-by-change. There is commentary in each commit message to explain every step.

The first 19 commits are shared with #1022 which lays the groundwork - please review that PR first. Once that is merged, viewing this PR will show only the later commits.

In the meantime, if you want to review the changes specific to this PR, start with [this commit](https://github.com/greatscottgadgets/hackrf/pull/982/commits/3eab937466bd04dc1e0f78c9c32cd798e6ff1594).

martinling:
I've just pushed a new version of the branch which uses spaces rather than tabs for aligning the pseudocode and cycle counts in `sgpio_m0.s`. This should ensure that the diffs don't show up ragged. Initial indenting is still done with a tab.

The first 19 commits are still shared with #1022. The new changes now start with [this commit](https://github.com/greatscottgadgets/hackrf/pull/982/commits/822fa5aed7623620484ba3efacd3fadc16ded677).

firmware/hackrf_usb/m0_state.c
```diff
@@ -1,6 +1,5 @@
 /*
```
mossmann, line 1: Can you please rename this to `usb_api_m0_state.c` for consistency with the other files that implement USB vendor requests?

firmware/hackrf_usb/sgpio_m0.s
```diff
@@ -1,108 +1,673 @@
...
+	new_mode .req r1
+
+	// Check count against threshold. If not a match, return to start of current loop.
+	ldr threshold, [state, #THRESHOLD]              // threshold = state.threshold          // 2
+	cmp count, threshold                            // if count != threshold:               // 1
+	bne \name\()_loop                               //      goto loop                       // 1 thru, 3 taken
+
+	// Otherwise, load and set new mode.
+	ldr new_mode, [state, #NEXT_MODE]               // new_mode = state.next_mode           // 2
+	str new_mode, [state, #MODE]                    // state.mode = new_mode                // 2
```
mossmann, line 343: I think there is a very low probability race condition where the M4 sets the mode, but then the M0 clobbers it immediately with the contents of NEXT_MODE.  This could be mitigated by having the M4 set NEXT_MODE before setting MODE; however, that could result in a different problem where the M0 follows the new NEXT_MODE in `jump_next_mode` before checking to see if the M4 has changed MODE.  In this case the new mode would be entered without resetting the stats.

firmware/hackrf_usb/usb_api_transceiver.c
```diff
@@ -262,29 +274,34 @@ void set_transceiver_mode(const transceiver_mode_t new_transceiver_mode) {
 		led_off(LED3);
 		led_on(LED2);
 		rf_path_set_direction(&rf_path, RF_PATH_DIRECTION_RX);
-		usb_bulk_buffer_tx = false;
+		set_m0_mode(M0_MODE_RX);
```
mossmann, line 277: It looks like this (and other mode changes) will result in the correct behavior only if the host has first asked us to set the mode to IDLE, but nothing enforces that. I feel like this should return an error to the host if the current mode does not support a direct transition to the requested mode. However, it might be better to accept whatever the host asks and forward the request to the M0 (with some changes to the M0 code to handle mode change requests (rather than having the M4 clobber the M0's mode directly).

firmware/hackrf_usb/sgpio_m0.s
```diff
@@ -1,108 +1,673 @@
...
+	bmi tx_zeros                                    // if buf_margin < 0: goto tx_zeros     // 1 thru, 3 taken
+
+	// Update buffer pointer.
+	update_buf_ptr                                  // update_buf_ptr()                     // 3
+
+	// At this point we know there is TX data available.
+	// If still in TX start mode, switch to TX run.
+	cmp mode, #MODE_TX_START                        // if mode != TX_START:                 // 1
+	bne tx_write                                    //      goto tx_write                   // 1 thru, 3 taken
+	mov mode, #MODE_TX_RUN                          // mode = TX_RUN                        // 1
```
mossmann, line 565: This could clobber a new mode that was just set by the M4.

firmware/hackrf_usb/sgpio_m0.s
```diff
@@ -1,108 +1,673 @@
...
+	ldr limit, [state, #SHORTFALL_LIMIT]            // limit = state.shortfall_limit        // 2
+	cmp limit, #0                                   // if limit == 0:                       // 1
+	beq \name\()_loop                               //      goto loop                       // 1 thru, 3 taken
+	cmp length, limit                               // if length < limit:                   // 1
+	blt \name\()_loop                               //      goto loop                       // 1 thru, 3 taken
+
+	// If so, reset mode to idle and return to idle loop.
+	mode .req r3
+	mov mode, #MODE_IDLE                            // mode = MODE_IDLE                     // 1
+	str mode, [state, #MODE]                        // state.mode = mode                    // 2
```
mossmann, line 402: This could clobber a new mode that was just set by the M4.

firmware/hackrf_usb/usb_api_transceiver.c
```diff
@@ -247,6 +250,15 @@ transceiver_mode_t transceiver_mode(void) {
 	return _transceiver_mode;
 }
 
+static void set_m0_mode(enum m0_mode mode) {
+	// All values of the threshold setting are valid. So setting a mode,
+	// it's necessary to set up a transition back to the same mode, for
+	// when the M0 counter wraps.
+	m0_state.mode = mode;
```
mossmann, line 257: In order to avoid race conditions, perhaps the M4 should never write to `mode` directly. There could be a separate `requested_mode` or maybe the existing `next_mode` could be used (either by setting `threshold` to a time in the near future or by adding a flag that indicates that the mode should be changed as soon as possible).

mossmann (changes requested):
This is fabulous work, wonderfully documented! I'm excited to get this merged soon, but I have requested changes, some of which will likely require discussion.

martinling:
Thanks @mossmann - I'd actually already started worrying about the races you mention. I was thinking along the same lines as you at first in terms of a fix: use separate variables for the mode requested by the M4, and the mode currently active on the M0, with each core only writing to its own variable.

But I'm not sure that change is actually sufficient.

Suppose that we split `mode` into `active_mode` (written only by the M0, read-only to the M4) and `requested_mode` (written only by the M4, read-only to the M0).

That does solve the races in which the M0 changes its mode autonomously and might clobber a requested change from the M4, e.g. during `TX_START` -> `TX_RUN`, transitions to `IDLE` at the shortfall limit, and scheduled changes to `next_mode`. So far so good.

But after getting to the bottom of #1042 (see [this commit](https://github.com/greatscottgadgets/hackrf/pull/1029/commits/7057235a145ec9bad7e3d6a240a250597214e16a) for the cause and fix), I'm thinking that we may need a similar solution here - i.e. give the M4's mode change requests a sequence number, so that the M0 can tell if it missed a request.

Otherwise, even if we change the loop exit conditions on the M0 from `mode == IDLE` to `active_mode != requested_mode`, there is a chance of the M0 missing a rapid sequence of requested mode changes such as `RX`->`IDLE`->`RX`.

That sequence can't safely be skipped over, since the `IDLE`->`RX` transition is supposed to have the side effect of resetting the stats and counters. If the M0 misses the two changes in the requested mode (which could happen especially at low sample rates where there is more time between SGPIO interrupts), then it will end up in the right mode but that side effect won't happen.

If we add a sequence number to those requests, then the M0 can compare the sequence number instead of the requested mode, and thus always tell if at least one new request was made. It can then transition `active_mode` to `IDLE`, go into the idle loop, and dispatch from there to whichever mode was actually requested.

For this to work though, writes and reads of the (`requested_mode`, `sequence_no`) pair need to be effectively atomic, because unlike in the #1042 fix, neither side can temporarily stop the other from running. The simplest solution would be to put the two in the same 32-bit word, e.g. set `m4_request = (sequence_no << 8) | requested_mode`. Then make the loop exit condition something like `m4_request != last_request`. But that would require either using up a register to keep the old value, or a extra load from memory.

I suppose another solution would be to have the M4' `set_m0_mode()` function block until the M0 has changed modes, i.e. do:

```c
    m0_state.requested_mode = new_mode
    while (m0_state.active_mode != new_mode);
```

That should ensure that the M0 can't miss a request, but it does mean that the M4 gets blocked until the M0 notices the request, which can take a while at low sample rates. We call that function from the `set_transceiver_mode` vendor request handler, which runs from the USB0 ISR, so I don't realy like having it block.

That said, after finding the races involved in #1042, I'm inclined to modify that request handler such that all it does is set a _requested_ transceiver mode, with the actual work of `set_transceiver_mode()` happening in the main loop in response to that.

It seems like we're converging on similar asynchronous control patterns being required for the host controlling the behaviour of the M4, and the M4 controlling the behaviour of the M0.

mossmann:
> I suppose another solution would be to have the M4's `set_m0_mode()` function block until the M0 has changed modes

I think this could work well, and I like the idea of moving the `set_transceiver_mode()` work out of the ISR.

Another way to minimize the effects of blocking on the M4 would be to wait for the previous mode before setting the new mode, e.g.:

```
while (m0_state.active_mode != m0_state.requested_mode);
m0_state.requested_mode = new_mode;
```

That way the time waiting for the M0 to change modes would overlap with the time between USB requests.

If we were to combine this with enforcement of passing through the IDLE mode, we could even have it block only while entering or exiting IDLE mode, not both:

```
if (new_mode != M0_MODE_IDLE) {
    m0_state.requested_mode = M0_MODE_IDLE;
    while (m0_state.active_mode != M0_MODE_IDLE;
}
m0_state.requested_mode = new_mode;
```

If the host is well-behaved, the transition to IDLE would have been previously requested, so there would be little or no blocking.

firmware/hackrf_usb/usb_api_transceiver.c
martinling: Fixed in cb3cfe8

firmware/hackrf_usb/sgpio_m0.s
martinling: Fixed in cb3cfe8

firmware/hackrf_usb/sgpio_m0.s
martinling: Fixed in cb3cfe8

firmware/hackrf_usb/usb_api_transceiver.c
martinling: Fixed in cb3cfe8

firmware/hackrf_usb/m0_state.c
martinling: Done

host/libhackrf/src/hackrf.c
```diff
@@ -943,6 +944,31 @@ int ADDCALL hackrf_rffc5071_write(hackrf_device* device, uint8_t register_number
...
+	USB_API_REQUIRED(device, 0x0106)
+	int result;
+
+	result = libusb_control_transfer(
+		device->usb_device,
+		LIBUSB_ENDPOINT_IN | LIBUSB_REQUEST_TYPE_VENDOR | LIBUSB_RECIPIENT_DEVICE,
+		HACKRF_VENDOR_REQUEST_GET_M0_STATE,
+		0,
+		0,
+		(unsigned char*)state,
```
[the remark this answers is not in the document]
martinling: Done the slightly less tedious way in ee00a1e.

gozu42:
(after getting side-tracked with #1045 ...) 

ee00a1e builds fine and passes my basic testing, both raspi and pc.

mossmann (approved):
Wonderful! I love how this turned out!

PR #985, opened 2021-11-08
sgpio_cpld_stream_rx_set_decimation() is no longer available

lyusupov:


miek (approved):
Yep, that was removed in 533f9ee33221981dedc5d55ec6becb9921121d35

mossmann:
Thank you for the clean up!

PR #990, opened 2021-11-10
Fix automatic rebuilding of M0 code

### LINKED ISSUE #693: Embedded M0 image isn't reliably rebuilt when sources change
Because `m0_bin.s` includes `${project_name}_m0.bin` with `.incbin`, CMake doesn't recognise the source dependency and doesn't rebuild `m0_bin.s.obj` when the .bin changes. The existing `add_dependencies` line doesn't solve this.

miek:
This uses `add_custom_command` to run a cmake script instead of using
`configure_file` directly, which allows for adding dependencies and
creates an ouput that can be depended on.

It also uses `add_custom_command` instead of `add_custom_target` to
create the M0 binary, since custom targets have no output file to depend
on.

Unfortunately, multiple targets cannot depend on the same generated file
or cmake tries to run the generation multiple times in parallel. So, the
.bin/.s generation is now duplicated for each target so they don't
conflict.

fixes #693

### DIFF
--- a/firmware/common/configure_file.cmake
+++ b/firmware/common/configure_file.cmake
@@ -1,0 +1,4 @@
+configure_file(
+    ${SRC}
+    ${DEST}
+)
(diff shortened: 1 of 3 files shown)


mossmann (approved):
Awesome! Works for me. Thank you!

PR #993, opened 2021-11-14
hackrf_sweep: simplify dwell time, bin width, and time stamps

### LINKED ISSUE #865: hackrf_sweep: sweep count issues
seeing this ... https://github.com/mossmann/hackrf/pull/856 
... reminds me sweep count is broken in at least one more way: 
when cranking up num_samples, each block for frequencies[0] is increasing sweep count by one. 
(== a single sweep with three data blocks per tuning is counted as three sweeps)

(and that probably causes problems with ifftw mode and anything else in this "is this a start of a sweep" code block too)

trivial fix would be to track the last_freq and only jump in there if current is freq[0] and last was not.

this would break sweep for a usecase that does not actualy retune, but i dont think that is even possible with interleaved mode currently and slightly silly to use hackrf_sweep in this case to begin with.

### LINKED ISSUE #951: Valid range of width for hackrf_sweep?
### Steps to reproduce
1.
2.
3.

### Expected behaviour
Able to error out if the width is outside of (unknown) spec while using hackrf_sweep utility. 

### Actual behaviour
Unknown

### Version information
*
Documentation

### Output
Documentation shows valid range of width. 
```

mossmann:
This removes support for longer dwell times at each frequency, something that I don't think has ever worked except perhaps in pre-release versions of hackrf_sweep.

fixes #865
fixes #951

### DIFF
--- a/host/hackrf-tools/src/hackrf_sweep.c
+++ b/host/hackrf-tools/src/hackrf_sweep.c
@@ -349,7 +349,7 @@ static void usage() {
 	fprintf(stderr, "\t[-p antenna_enable] # Antenna port power, 1=Enable, 0=Disable\n");
 	fprintf(stderr, "\t[-l gain_db] # RX LNA (IF) gain, 0-40dB, 8dB steps\n");
 	fprintf(stderr, "\t[-g gain_db] # RX VGA (baseband) gain, 0-62dB, 2dB steps\n");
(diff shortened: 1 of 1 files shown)


gozu42:
LGTM
removes mostly confusing code for two "features" that never really did what they were meant to, 
and improves documentation for a related feature left.
:+1:

PR #1004, opened 2021-12-01
fix stream glitch on rate change

### LINKED ISSUE #1003: Samplerate changes cause IQ glitch in stream
### Steps to reproduce
1. constantly change the samplerate of a running stream between two different rates
2. after usually between 20 - 100 ratechanges the "IQIQIQ .." data gets shifted to "QIQIQIQI .." samples (a one byte shift in the stream)
3. after the same random duration it glitches back into a working state
4. if no further rate change happens the misalignment persists 

### Expected behaviour
this should not happen

### Actual behaviour
the same physical channel gets mapped to I in one run and Q in the next.

### Version information
(16,18,20.04) Ubuntu
(2018, 2021, latest firmware/library)

### Output
Screenshot from 2021-11-30 17-20-02

in the setup i had lose cables connected to the I+ and I- pads, rendering this channel noisy while the Q channel was shorted. This shows how the same physical channel turned form Q to I.

this is exactly the same issue as #439

gibbi:
this changes fixes #1003 by disabling cpld processing while changing the samplingrate

### DIFF
--- a/firmware/common/hackrf_core.c
+++ b/firmware/common/hackrf_core.c
@@ -23,6 +23,7 @@
 
 #include "hackrf_core.h"
 #include "hackrf_ui.h"
+#include "sgpio.h"
 #include "si5351c.h"
 #include "spi_ssp.h"
 #include "max2837.h"
@@ -373,6 +374,12 @@ bool sample_rate_frac_set(uint32_t rate_num, uint32_t rate_denom)
 		}
 	}
 
+	bool streaming = sgpio_cpld_stream_is_enabled(&sgpio_config);
+	
+	if (streaming) {
+		sgpio_cpld_stream_disable(&sgpio_config);	
+	}
+	
 	/* Can we enable integer mode ? */
 	if (a & 0x1 || b)
 		si5351c_set_int_mode(&clock_gen, 0, 0);
@@ -392,6 +399,10 @@ bool sample_rate_frac_set(uint32_t rate_num, uint32_t rate_denom)
 
 	/* MS0/CLK2 is the source for SGPIO (CODEC_X2_CLK) */
 	si5351c_configure_multisynth(&clock_gen, 2, 0, 0, 0, 0);//p1 doesn't matter
+
+	if (streaming) {
+		sgpio_cpld_stream_enable(&sgpio_config);
+	}
 
 	return true;
 }


martinling (approved):
I'm not that familiar with the details of the CPLD yet, but I've traced the effect of this through the schematic and the HDL to confirm how this works, and it makes sense to me as follows:

Setting `SGPIO_GPIO_OUTREG |= (1L << 10)` sets the SGPIO10 output high.

That's pin 61 on the LPC4320, on the `SGPIO_10` net, which is connected via R88 to net `BANK2F3M8` which connects to pin 76 of the CPLD.

In the CPLD project, `top.ucf` assign that pin as `NET "HOST_DISABLE"  LOC = "P76"`.

Then in `top.vhd` that's connected with `host_data_enable_i <= not HOST_DISABLE`.

Which brings us to the relevant bit of VHDL:
```vhdl
    process (host_data_enable_i, host_sync_i)
    begin
        host_sync_o <= host_data_enable_i;
        if host_data_enable_i = '1' then
            if rising_edge(host_sync_i) then
                host_sync_latched <= host_sync_i;
            end if;
        else
            host_sync_latched <= '0';
        end if;
    end process;
```
So the effect should be to deassert the `host_data_enable_i` signal whilst the sample rate is changed, causing the CPLD to resync after the change is completed.

Provided that this has been tested and shown to fix the desync seen in #1003, I don't see any problem with it.

The `sample_rate_frac_set` function is only called during the USB vendor request to set the sample rate, and for our common use cases that won't be called during streaming, so I don't see much obvious opportunity for side effects.

firmware/common/hackrf_core.c
```diff
@@ -373,6 +373,8 @@ bool sample_rate_frac_set(uint32_t rate_num, uint32_t rate_denom)
 		}
 	}
 
+	SGPIO_GPIO_OUTREG |= (1L << 10); // disable cpld processing
```
miek, line 376: This should call `sgpio_cpld_stream_disable` instead of poking at the register directly.

firmware/common/hackrf_core.c
```diff
@@ -393,6 +395,8 @@ bool sample_rate_frac_set(uint32_t rate_num, uint32_t rate_denom)
 	/* MS0/CLK2 is the source for SGPIO (CODEC_X2_CLK) */
 	si5351c_configure_multisynth(&clock_gen, 2, 0, 0, 0, 0);//p1 doesn't matter
 
+	SGPIO_GPIO_OUTREG &= ~(1L << 10); // enable cpld processing
```
miek, line 398: As above, with `sgpio_cpld_stream_enable`.

miek (changes requested):
Thanks for the PR! I was able to reproduce the original issue, and this PR fixes the problem for me too.

I've got a couple of suggestions for slight changes before merging:
1. I added some inline comments pointing to functions to call instead of modifying the registers directly
2. This should check that streaming is currently enabled by calling `sgpio_cpld_stream_is_enabled`, before calling `sgpio_cpld_stream_disable` and `sgpio_cpld_stream_enable`.

firmware/common/hackrf_core.c
gibbi: fixed, i prefer not to check if it was disabled before .. i like to only add ifs when they would make a difference somehow.

firmware/common/hackrf_core.c
gibbi: fixed

firmware/common/hackrf_core.c
```diff
@@ -393,6 +396,8 @@ bool sample_rate_frac_set(uint32_t rate_num, uint32_t rate_denom)
 	/* MS0/CLK2 is the source for SGPIO (CODEC_X2_CLK) */
 	si5351c_configure_multisynth(&clock_gen, 2, 0, 0, 0, 0);//p1 doesn't matter
 
+	sgpio_cpld_stream_enable(&sgpio_config);
```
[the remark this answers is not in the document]
gibbi: right - i did not think about that :-) fixed

miek (approved):

martinling:
Thanks for the fix @gibbi !

PR #1023, opened 2021-12-22
Fix double hyphen (--) rendered as en dash

icblnk:
Enclosed `hackrf_debug` tool invocation into double back quotes (``) so that when it's rendered to HTML double hyphen is not converted to en dash.


straithe (approved):
Great catch! Thank you for the contribution.

Issue #916, opened 2021-07-12
Repeated Start/Stop RX Produces Error


AlexFWulff:
I think that this is still a problem on the most recent release (2021.03.1). I wrote a C program that can reproduce an error where, after a few repeated start rx/stop rx calls in quick succession, the TX light stays on and the rx callback is never triggered. 

Here is my hackrf_info output (platform is RPi 3 raspbian):

```
hackrf_info version: 2021.03.1
libhackrf version: unknown (0.5)
Found HackRF
Index: 0
Serial number: 0000000000000000088869dc24397f1b
Board ID Number: 2 (HackRF One)
Firmware Version: 2021.03.1 (API:1.04)
Part ID Number: 0xa000cb3c 0x0041474f
```

libhackrf is not installed globally - instead I'm manually linking in the compiled libhackrf 2021.03.1 binary.

Check out [this gist](https://gist.github.com/AlexFWulff/7d6e75aaffb7215c2a175a2992db70e1) for the program I'm using. 

Below is the expected output. The rx should start, maybe get a few cycles before the rx callback is called, and then should restart again indefinitely.

```
Start
Not started yet
Valid length: 262144
Valid length: 262144
Valid length: 262144
Finally started
Start
Not started yet
Valid length: 262144
Valid length: 262144
Finally started
Start
Valid length: 262144
....
```

Below is the output I get after letting the program run for about 10 seconds (sometimes more, sometimes less). Eventually, the callback is never triggered and the program gets stuck on "Not started yet" forever.

```
Finally started
Start
Not started yet
Valid length: 262144
Valid length: 262144
Valid length: 262144
Valid length: 262144
Valid length: 262144
... [repeats a few times]
Finally started
Start
Not started yet
Not started yet
Not started yet
Not started yet
Not started yet
... [continues forever]
```

Is there a way to fix this to allow for start_rx and stop_rx to be called in quick succession?

_Originally posted by AlexFWulff in https://github.com/mossmann/hackrf/issues/394#issuecomment-878362478_

straithe:
Have you tried reproducing this issue on a Linux desktop or laptop that is not a Raspberry Pi?

AlexFWulff:
Yes - I have tried this on macOS natively and in a Debian Virtualbox instance.

xaki23:
i can repro this with the supplied example code on both a raspi4 and fedora-vm.
tried some variants, but cant make much sense of it yet.
this is meant as a rough collection of observations.

[...]

Issue #883, opened 2021-04-24
Freeze in Gqrx with hackRF 2021-03-1 driver version


Potomac:
### Steps to reproduce
1. Install Gqrx software, 2.14.4 version
1. Use hackRF device (february 2014 PCB version)
1. Update firmware to 2021-03 version
1. Update hackRF driver to 2021-03-1 version
1. Use linux (archlinux 64 bits)
1. Run Gqrx
1. Switch several times FM to AM (and vice-versa)
1. Randomly Gqrx will freeze

### Expected behaviour
Gqrx should no randomly freeze during operation

### Actual behaviour
Randomly, when switching frequency mode (AM, FM, LSB etc...) then Gqrx can freeze, I hear no sound from hackRF, it can happen 30% of the time when switching frequency mode,

if I use 2018 hackrf driver version then there is no freeze, no bugs, all is Ok with Gqrx,
my hackRF is a 2014 PCB version, firmware has been updated to 2021-03

### Version information
archlinux 64 bits

hackrf_info version: 2021.03.1
libhackrf version: 2021.03.1 (0.6)
Board ID Number: 2 (HackRF One)
Firmware Version: 2021.03.1 (API:1.04)

Gqrx 2.14.4

Potomac:
If I use the official appimage of Gqrx (last version, same as archlinux package : 2.14.4) then there is no freeze, no bug with hackRF 2021-03-1 driver :
https://github.com/csete/gqrx/releases

Potomac:
But the appimage can also contain hackRF lib in 2018-01 version, that can explain why there is no freeze.

straithe:
Have you tried your HackRF with other software on Arch?

Potomac:
Yes, I tried 2 softwares : gnu radio and hackTV, both work with hackRF 2021-03-01.

Gqrx freezes randomly with hackRF 2021-03-01 when switching modes (AM, FM, LSB), and sometimes when I start/stop hackRF with the menu "file -> start dsp", and "File -> stop dsp", it seems that the driver in 2021-03 version may hang the device.

Potomac:
@straithe : do you have a test script/binary that can check if all features, functions of hackRF driver 2021-03-01 are Ok ?

I use 2 commands for checking transmission and reception and it seems ok : 
hackrf_transfer -t /dev/zero
hackrf_transfer -r /dev/null -s 21500000

[...]

PR #1029, opened 2021-12-31
Fixes for concurrency bugs in start/stop operations

martinling:
Fixes bug #916.

Previously, there was a race which could lead to a transfer being left active after `cancel_transfers()` completed. This would then cause the next `prepare_transfers()` call to fail, because `libusb_submit_transfer()` would return an error due to the transfer already being in use.

The sequence of events that could cause this was:

1. Main thread calls `hackrf_stop_rx()`, which calls `cancel_transfers()`, which iterates through the 4 transfers in use and cancels them one by one with `libusb_cancel_transfer()`.
1. During this time, a transfer is completed. The transfer thread calls `hackrf_libusb_transfer_callback()`, which handles the data and then calls `libusb_submit_transfer()` to resubmit that transfer.
1. Now, `cancel_transfers()` and `hackrf_stop_rx()` are completed but one transfer is still active.
1. The next `hackrf_start_rx()` call fails, because `prepare_transfers()` tries to submit a transfer which is already in use.

To fix this, we add a lock which must be held to either cancel transfers or restart them. This ensures that only one of these actions can happen for a given transfer; it's no longer possible for a transfer to be cancelled and then immediately restarted.

With this change, I can now run the test program from #916 without failures.

### DIFF
--- a/firmware/hackrf_usb/hackrf_usb.c
+++ b/firmware/hackrf_usb/hackrf_usb.c
@@ -263,15 +263,29 @@ int main(void) {
 	operacake_init(operacake_allow_gpio);
 
 	while(true) {
-		switch (transceiver_mode()) {
+
+		// Briefly disable USB interrupt so that we can
+		// atomically retrieve both the transceiver mode
+		// and the mode change sequence number. They are
+		// changed together by the set_transceiver_mode
+		// vendor request handler.
+
+		nvic_disable_irq(NVIC_USB0_IRQ);
+
+		transceiver_mode_t mode = transceiver_mode();
+		uint32_t seq = transceiver_mode_seq();
+
+		nvic_enable_irq(NVIC_USB0_IRQ);
+
+		switch (mode) {
 		case TRANSCEIVER_MODE_RX:
-			rx_mode();
+			rx_mode(seq);
 			break;
 		case TRANSCEIVER_MODE_TX:
-			tx_mode();
+			tx_mode(seq);
 			break;
 		case TRANSCEIVER_MODE_RX_SWEEP:
-			sweep_mode();
+			sweep_mode(seq);
 			break;
 		case TRANSCEIVER_MODE_CPLD_UPDATE:
 			cpld_update();
--- a/firmware/hackrf_usb/usb_api_sweep.c
+++ b/firmware/hackrf_usb/usb_api_sweep.c
@@ -87,7 +87,7 @@ usb_request_status_t usb_vendor_request_init_sweep(
 	return USB_REQUEST_STATUS_OK;
 }
 
-void sweep_mode(void) {
+void sweep_mode(uint32_t seq) {
 	unsigned int blocks_queued = 0;
 	unsigned int phase = 1;
 	bool odd = true;
@@ -98,7 +98,7 @@ void sweep_mode(void) {
 
 	baseband_streaming_enable(&sgpio_config);
 
-	while (TRANSCEIVER_MODE_RX_SWEEP == transceiver_mode()) {
+	while (transceiver_mode_seq() == seq) {
 		// Set up IN transfer of buffer 0.
 		if ( m0_state.offset >= 16384 && phase == 1) {
 			transfer = true;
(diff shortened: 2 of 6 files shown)


martinling:
I've added a second commit to fix an orthogonal but closely related issue.

After cancelling transfers, we need to wait for all cancellation handling to complete before trying to do anything new with the same transfers.
    
Calling `libusb_cancel_transfer()` only starts the cancellation of a transfer. The process is not complete until the transfer callback has been called with status `LIBUSB_TRANSFER_CANCELLED`.

If `hackrf_start_rx()` is called soon after `hackrf_stop_rx()`, `prepare_transfers()` may be called before the previous cancellations are completed, resulting in a `LIBUSB_ERROR_BUSY` when a transfer is reused with `libusb_submit_transfer()`.
    
To prevent this happening, I've made the transfer thread keep track of which transfers have finished (either by completion, or cancellation), and made `cancel_transfers()` wait until all transfers are finished before returning.
    
This is implemented using a pthread condition variable which is signalled from the transfer thread and waited on by `cancel_transfers()`.

With this change, I can no longer reproduce the failure to restart RX seen in #883.

martinling:
There's a recent [thread](https://lists.osmocom.org/hyperkitty/list/osmocom-sdr@lists.osmocom.org/thread/7TKOAOP5KNQBBCDJRV7EI67AFKEDTNCP/) on the osmocom-sdr list from Jasper van den Eshof, who has been fixing more or less exactly the same issue in librtlsdr.

Looking at the [librtlsdr code](https://github.com/steve-m/librtlsdr/blob/master/src/librtlsdr.c) suggests a simpler approach that could be used to orchestrate the cancellations, which would eliminate the need for the locking introduced in https://github.com/greatscottgadgets/hackrf/pull/1029/commits/b346790e2da866fd79e41393f3ddf2db6eb25102.

- In `hackrf_stop_{rx|tx}()`, set a flag to request cancellation, then call `libusb_interrupt_event_handler()`.
- In `transfer_threadproc`, use `libusb_handle_events_timeout_completed()` instead of `libusb_handle_events_timeout()`, and pass the flag set in `hackrf_stop_{rx|tx}` as the `completed` argument. This function will then return when requested.
- Once `libusb_handle_events_timeout_completed()` has completed, do the `cancel_transfers()` work in the transfer thread. The mutex between `cancel_transfers()` and `hackrf_libusb_transfer_callback()` would then not be necessary.
- After cancelling transfers, loop on `libusb_handle_events_timeout_completed()` until all cancellation callbacks have completed. Then signal `hackrf_stop_{rx|tx}` as before.

martinling:
Another useful observation from the discussion with Jasper - the individual `transfer_finished` flags, added in my second commit, could be more simply replaced with a count of active transfers, which is incremented on submission and decremented on completion or cancellation.

I'll work on making both improvements.

martinling:
I've added a third commit which fixes the firmware-side issue #1042.

With the combination of all three changes on this PR, I can no longer reproduce any problems with repeated start/stop of HackRF.

I plan to do some further work to simplify the solutions, but for anyone having problems this branch should work as-is.

gozu42:
can confirm no longer seeing any restart-hangs with 7057235a on pc+raspi with test app from https://github.com/greatscottgadgets/hackrf/issues/916 or gqrx ctrl-d stresstest.

martinling:
Since this PR has been tested & confirmed to fix #883, #916 and #1042, I propose we get it merged as-is without further changes, and we can look at making the simplifications discussed in [this comment](https://github.com/greatscottgadgets/hackrf/pull/1029#issuecomment-1017639509) and [this comment](https://github.com/greatscottgadgets/hackrf/pull/1029#issuecomment-1019592611) in a separate PR later.

mossmann:
Outstanding troubleshooting and solutions!

martinling:
The simplification discussed in [this comment](https://github.com/greatscottgadgets/hackrf/pull/1029#issuecomment-1019592611) is now in PR #1071.

I had a go at the the other proposed change outlined in [this comment](https://github.com/greatscottgadgets/hackrf/pull/1029#issuecomment-1017639509), but decided not to go ahead with it. Although it would eliminate some locking, it would just add different complications instead. It would detract from the simplicity of the transfer thread, which is currently just a trivial libusb event handling loop.

PR #1033, opened 2022-01-10
Add project mention for the APRS tracker

rgerganov:
APRS tracker which uses the HTML Geolocation API to retrieve the
device's position and WebUSB to communicate with the HackRF. The source
code is available at https://github.com/rgerganov/aprs-sdr


straithe (changes requested):
This would be better listed in ``software_support.rst`` and not ``hackrf_projects_mentions.rst``. Can you please make that change?

rgerganov:
> This would be better listed in `software_support.rst` and not `hackrf_projects_mentions.rst`. Can you please make that change?

Done

straithe:
Looks great! Thank you for the addition.

PR #1034, opened 2022-01-10
rad1o UI support

schneider42:
This adds support for the rad1o display.

Here is an example of how it looks like:
PXL!UNITO-UNDERSCORE!20220110!UNITO-UNDERSCORE!223104439

The display and rendering code is from the original rad1o firmware and was used for quite a bit more. I've tried to remove as much unneeded functionality as possible and did a rough cleanup of the code. Please let me know if you would like a different file organization / code style / function naming / header format etc...


mossmann (approved):
Wonderful! Thank you, schneider!

Issue #1042, opened 2022-01-29
Firmware USB streaming can stall after the first 16K USB transfer following RX restart


martinling:
This has been very hard to pin down, but I think I've now proven it.

In both #883 and #916, we saw host-side problems that caused a rapid RX stop/start sequence to fail, due to concurrency bugs in our usage of libusb. I believe I have now fixed all these host-side issues in #1029.

However, there was still a failure mode being seen in which no libusb errors occured, yet the RX data flow failed to restart. The RX LED was on when this happened, indicating that the transceiver mode change had been processed on the device. The libusb event handling was proceeding correctly, but a bulk transfer was never filled.

I've now succeeded in using LUNA to catch this happening on the wire. A pcap file, trimmed to just the relevant section, is attached: [clip.pcap.zip](https://github.com/greatscottgadgets/hackrf/files/7963478/clip.pcap.zip)

After the vendor requests to change the transceiver mode to `OFF`, and then to `RX`, only a single 16KB transfer from the firmware is completed. The host continues to send `IN` tokens to request data from the endpoint, but the device only responds with `NAK`.

From the host side, it appears as if no data at all is received, because we submit libusb transfers for 256KB of data, with no timeout.

I have no idea what's going wrong on the firmware side yet, but I do know that the symptoms are reproducible both with and without the changes in #982.

Issue #1019, opened 2021-12-13
Faster and cleaner exit from `hackrf_transfer`


mossmann:
`hackrf_transfer` could use some improvements similar to those made in `hackrf_sweep` over the past year.

* Check for completion more often than once per second (c47f2b1, 5361b3a).
* Remove extra calls to `hackrf_stop_rx()` and `hackrf_stop_rx()` when `hackrf_close()` is about to be called anyway (4c46fc7).
* Maybe jump to exit earlier, perhaps when the `-n` option is used (ab4498f).

PR #1045, opened 2022-02-08
Move transceiver mode changes out of USB ISR.

martinling:
This is a defensive change to make the transceiver code easier to reason about, and to avoid the possibility of races such as that seen in #1042.

Previously, `set_transceiver_mode()` was called in the vendor request handler for the `SET_TRANSCEIVER_MODE` request, as well in the callback for a USB configuration change. Both these calls are made from the USB0 ISR, so could interrupt the `rx_mode()`, `tx_mode()` and `sweep_mode()` functions at any point. It was hard to tell if this was safe.

Instead, `set_transceiver_mode()` has been removed, and its work is split into three parts:

- `request_transceiver_mode()`, which is safe to call from ISR context. All this function does is update the requested mode and increment a sequence number. This builds on work already done in PR #1029, but the interface has been simplified to use a shared volatile structure.

- `transceiver_startup()`, which transitions the transceiver from an idle state to the configuration required for a specific mode, including setting up the RF path, configuring the M0, adjusting LEDs and UI etc.

- `transceiver_shutdown()`, which transitions the transceiver back to an idle state.

The `*_mode()` functions that implement the transceiver modes now call `transceiver_startup()` before starting work, and `transceiver_shutdown()` before returning, and all this happens in the main thread of execution.

As such, it is now guaranteed that all the steps involved happen in a consistent order, with the transceiver starting from an idle state, and being returned to an idle state before control returns to the main loop.

For consistency of interface, an `off_mode()` function has been added to implement the behaviour of the OFF transceiver mode. Since the transceiver is already guaranteed to be in an idle state when this is called, the only work required is to set the UI mode and wait for a new mode request.

### DIFF
--- a/firmware/hackrf_usb/hackrf_usb.c
+++ b/firmware/hackrf_usb/hackrf_usb.c
@@ -146,7 +146,7 @@ void usb_configuration_changed(
 	usb_device_t* const device
 ) {
 	/* Reset transceiver to idle state until other commands are received */
-	set_transceiver_mode(TRANSCEIVER_MODE_OFF);
+	request_transceiver_mode(TRANSCEIVER_MODE_OFF);
 	if( device->configuration->number == 1 ) {
 		// transceiver configuration
 		led_on(LED1);
@@ -263,29 +263,30 @@ int main(void) {
 	operacake_init(operacake_allow_gpio);
 
 	while(true) {
+		transceiver_request_t request;
 
 		// Briefly disable USB interrupt so that we can
 		// atomically retrieve both the transceiver mode
 		// and the mode change sequence number. They are
-		// changed together by the set_transceiver_mode
-		// vendor request handler.
+		// changed together by request_transceiver_mode()
+		// called from the USB ISR.
 
 		nvic_disable_irq(NVIC_USB0_IRQ);
-
-		transceiver_mode_t mode = transceiver_mode();
-		uint32_t seq = transceiver_mode_seq();
-
+		request = transceiver_request;
 		nvic_enable_irq(NVIC_USB0_IRQ);
 
-		switch (mode) {
+		switch (request.mode) {
+		case TRANSCEIVER_MODE_OFF:
+			off_mode(request.seq);
+			break;
 		case TRANSCEIVER_MODE_RX:
-			rx_mode(seq);
+			rx_mode(request.seq);
 			break;
 		case TRANSCEIVER_MODE_TX:
-			tx_mode(seq);
+			tx_mode(request.seq);
 			break;
 		case TRANSCEIVER_MODE_RX_SWEEP:
-			sweep_mode(seq);
+			sweep_mode(request.seq);
 			break;
 		case TRANSCEIVER_MODE_CPLD_UPDATE:
 			cpld_update();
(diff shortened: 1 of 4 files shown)


martinling:
This PR is in draft, because it's currently not reliable. Sometimes, a `hackrf_transfer` run will leave the TX or RX LED on, and subsequent runs will fail. However, running the `hackrf_loop` test program from #916 always works.

I haven't figured out why yet.

gozu42:
i dont have a clear answer to the "why" either, but i will try to summarize my observations.

* hackrf_transfer with -n keeps the rx led on here. subsequent runs work, and also keep the rx led on.
* hackrf_transfer with ctrl-c switches the rx led off.
* hackrf_sweep with -N switches the rx led off.

while doing that, i noticed ... 
* hackrf_transfer with -n stops receiving when hitting the limit, but only exits the main loop when hitting the "didnt receive anything for >1sec" condition.
* which is related to #1019 not being implemented.

after transplanting the "have rx_callback set applevel do_exit when reaching -n limit" code from hackrf_sweep to hackrf_transfer (as a crude/partial approximation of #1019 changes)
* hackrf_transfer with -n now exits without the timeout and switches the rx led off.
* one change imported with the hackrf_sweep code there is that rx_callback now returns 0 instead of -1 when reaching the limit.
* if i keep the do_exit handling, but change it to return -1, the rx led stays on again.

so my current summary here would be "rx_callback returning -1 somehow makes stop_rx calls return success, but doesnt actualy seem to run set_transceiver_mode(OFF) successfully, because there is no real way for the led to stay on if that happened".

martinling:
I concluded that the problem must be to do with the `usb_endpoint_flush` calls, because everything else that was previously done by `set_transceiver_mode` seemed unable to affect the flow of execution.

My first idea was to disable the USB ISR around the `usb_endpoint_flush` calls in `transceiver_shutdown`, suspecting a concurrency issue between these calls and the ISR. But this change had no effect.

What _did_ work is to keep the `usb_endpoint_flush` calls in the USB ISR. It's not clear to me why this is necessary, but it does seem to fix the problem reliably, so I've added a commit that does this.

I have kept the `usb_endpoint_flush` calls in `transceiver_shutdown()` as well, because otherwise there would be a race where the USB ISR could return just before `rx_mode` queues a new transfer, and ends up leaving one in flight.

martinling:
@gozu42 thanks for catching that, I've reproduced here and it looks like it was just a problem with the rebase, not with either PR alone.

In the sweep loop there were a couple of instances of:
```c
    while (m0_state.active_mode != M0_MODE_WAIT)
        if (transceiver_request.seq != seq)
            return;
```

But with the changes here in #1045 those need to be a `break`, not a `return`, so that the `transceiver_shutdown()` at the end gets run. I've updated the `bug-180-rebase` branch accordingly now.

martinling:
And fixed again, because of course a `break` there only gets out of that inner while loop, so it's now a `goto` leading to the `transceiver_shutdown()` call.

gozu42:
bug-180-rebase at 779483b9 and usb-isr at 838ad072 now pass all my current standard usecase tests.

PR #1069, opened 2022-03-18
Overhaul handling of transfer errors and use of streaming flag.

### LINKED ISSUE #1068: libhackrf hangs if HackRF is disconnected
### Steps to reproduce
1. Start `hackrf_transfer -r /dev/null`
2. Unplug HackRF
### Expected behaviour
Program exits with an appropriate error.
### Actual behaviour
Program hangs, and can only be killed with `SIGKILL`.
### Background
I decided to test this after writing [this comment](https://github.com/greatscottgadgets/hackrf/pull/1057/#issuecomment-1069146134), because disconnection was the obvious case for when we would need to handle errors from libusb.

It looks like I broke this in #1029.

Running with gdb shows that the hang happens when `hackrf_stop_rx()` calls `cancel_transfers()`, which calls `pthread_cond_wait` to wait for cancellations to complete. But the transfer thread is already dead, because `hackrf_libusb_transfer_callback` saw an error and called `request_exit`. So the condition is never set.

We need to rethink the error handling path to work with the other changes. I'll work on that.

martinling:
This PR ensures correct behaviour during shutdown, and reviews and simplifies usage of the `streaming` and `do_exit` flags. The changes are as follows:

- The transfer callback no longer calls `request_exit()` to end the transfer thread if an error occurs. It was no longer safe to do this after the changes in #1029, because we need the transfer thread to keep running, to handle cancellations and signal when these are complete. Instead, if it sees an error, the callback just records the transfer as finished, and clears the `streaming` flag. This fixes #1068.
- The transfer callback will no longer call the user RX callback, or submit further transfers, once the `streaming` flag is cleared. This ensures that:
  - Once the RX callback returns nonzero to indicate completion, it is not called again with further in-flight transfers.
  - After any libusb error occurs, no further RX callbacks are made and no more transfers are submitted.
- Review, correct and simplify all the cases where we set or clear `streaming` and `do_exit`. The result is:
  - The `streaming` flag is:
    - Set in `prepare_transfers`, which is called via `prepare_setup_transfers` by:
      - `hackrf_start_tx`
      - `hackrf_start_rx`
      - `hackrf_start_rx_sweep`
    - Cleared in `cancel_transfers`, which is called by:
      - `hackrf_stop_tx`
      - `hackrf_stop_rx`
      - `hackrf_close` via `kill_transfer_thread`
    - Cleared in `transfer_finished`, which is called by `hackrf_libusb_transfer_callback` if any transfer finishes without being successfully resubmitted.
    - Cleared in `transfer_threadproc` if `libusb_handle_events_timeout` returns an error.
  - The `do_exit` flag is set only in `kill_transfer_thread`, which is called only from `hackrf_close`. The flag is cleared again after joining the thread.

### DIFF
--- a/host/libhackrf/src/hackrf.c
+++ b/host/libhackrf/src/hackrf.c
@@ -181,11 +181,6 @@ static int create_transfer_thread(hackrf_device* device);
 static libusb_context* g_libusb_context = NULL;
 int last_libusb_error = LIBUSB_SUCCESS;
 
-static void request_exit(hackrf_device* device)
-{
-	device->do_exit = true;
-}
-
 /*
  * Check if the transfers are setup and owned by libusb.
  *
@@ -1781,10 +1776,10 @@ static int kill_transfer_thread(hackrf_device* device)
 		 * thread has handled all completion callbacks.
 		 */
 		cancel_transfers(device);
-		/*
-		 * Now call request_exit() to halt the main loop.
-		 */
-		request_exit(device);
+
+		// Set flag to tell the thread to exit.
+		device->do_exit = true;
+
 		/*
 		 * Interrupt the event handling thread instead of
 		 * waiting for timeout.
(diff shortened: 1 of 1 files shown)


metayan:
Tested on Linux (funtoo) and Darwin (10.15.7).
Nice.

mossmann (approved):
Wonderful!

PR #1070, opened 2022-03-18
Remove unnecessary delays on stop, and duplicated stop commands on close.

martinling:
This is stacked on #1069 to avoid conflicts, so until that's merged, start reviewing from commit [`958c742`](https://github.com/greatscottgadgets/hackrf/pull/1070/commits/958c7421890c62436e198547473ccfcd3d838c46).

The changes in this PR are:

- Remove the 10ms delays that were added to `hackrf_stop_rx_cmd` and `hackrf_stop_tx_cmd` in PR #805.
  - These delays were a workaround to ensure that their calling functions did not return until transfer cancellations had completed. That problem is now fixed properly in #1029, so the delays are no longer necessary.
- Merge `hackrf_stop_rx_cmd` and `hackrf_stop_tx_cmd`, since they do the same thing: set transceiver mode to OFF.
- Remove the duplicate use of that command in `hackrf_close`.

### DIFF
--- a/host/libhackrf/src/hackrf.c
+++ b/host/libhackrf/src/hackrf.c
@@ -1947,22 +1947,20 @@ int ADDCALL hackrf_stop_tx(hackrf_device* device)
 
 int ADDCALL hackrf_close(hackrf_device* device)
 {
-	int result1, result2, result3;
+	int result1, result2;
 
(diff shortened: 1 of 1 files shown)


mossmann (approved):
Looks great!

PR #1071, opened 2022-03-20
Simplify logic for tracking finished transfers

martinling:
This is a followup cleanup to #1029.

This PR simplifies the logic for tracking finished transfers, by removing the per-transfer `transfer_finished` flags and the overall `all_finished` flag, and replacing them with a count of `active_transfers`. 

From a concurrency point of view everything works as before, except that the pthread condition variable is now associated with the assertion `active_transfers == 0` rather than `all_finished`. The associated lock must therefore be held to decrement that count to zero, and the condition is signalled when this happens.

I proposed this simplification in [this comment](https://github.com/greatscottgadgets/hackrf/pull/1029#issuecomment-1019592611) on PR #1029, but left it out of that PR in order to get a working fix merged ASAP.


martinling:
Rebased for easier viewing now that #1069 and 1070 are merged.

martinling:
Added a commit to make it easier to reason about the locking.

mossmann (approved):
Wonderful! I appreciate the easier-to-read locking.

PR #1090, opened 2022-04-19
Bugfix in sweep mode when multiple blocks have to be read for each frequency

veloman-yunkan:
This PR fixes a bug introduced by #982 (more precisely, by commit 9f79a16b261136d89d4baeff958168b7af8a71e6). The said change overlooked that in sweep mode the count of samples to be read for each frequency may be set to more than one block (tracked by the variable `dwell_blocks`, which became unused).

The diff is much-much simpler if whitespace changes are ignored.

### DIFF
--- a/firmware/hackrf_usb/usb_api_sweep.c
+++ b/firmware/hackrf_usb/usb_api_sweep.c
@@ -121,6 +121,7 @@ void sweep_mode(uint32_t seq) {
 	//
 	// 7. Process repeats from step 1.
 
+	unsigned int blocks_queued = 0;
 	unsigned int phase = 0;
 	bool odd = true;
 	uint16_t range = 0;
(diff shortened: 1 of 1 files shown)


martinling (approved):
Hi @veloman-yunkan, sorry it's taken me a while to look at this.

Thanks for catching my error. I hadn't realised that the dwell blocks feature was still in use when I rewrote the sweep code - I think I must have assumed that it was a deprecated feature like the non-interleaved sweep style.

I assume you're using some software which calls libhackrf's `hackrf_init_sweep` with custom parameters, since the `hackrf_sweep` command line tool only ever passes 16384 as `num_bytes`.

The fix looks fine to me. For testing purposes I modified `hackrf_sweep.c` with [this patch](https://github.com/greatscottgadgets/hackrf/files/8798673/hackrf_sweep_dwell_4x.patch.txt) to use 4 blocks at each tuning step rather than 1. Everything seemed to work fine, with the expected 4x reduction in sweep rate.

We might want to consider adding a command line argument to `hackrf_sweep` to support this feature.

mossmann:
Thank you for the contribution!

PR #1103, opened 2022-06-05
fixed firmware manual build docs

dizcza:
It ain't much but it's honest work.

### DIFF
--- a/firmware/README
+++ b/firmware/README
@@ -31,6 +31,12 @@ $ hackrf_spiflash -w hackrf_usb.bin
 If you have a Jawbreaker, add -DBOARD=JAWBREAKER to the cmake command.
 If you have a rad1o, use -DBOARD=RAD1O instead.
 
+If you get the "`arm-none-eabi-gcc` is not a full path and was not found in the PATH"
+error during `cmake ..`, install the package with `sudo apt-get install gcc-arm-none-eabi`.
+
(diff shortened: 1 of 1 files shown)


firmware/README
```diff
@@ -31,6 +31,12 @@ $ hackrf_spiflash -w hackrf_usb.bin
 If you have a Jawbreaker, add -DBOARD=JAWBREAKER to the cmake command.
 If you have a rad1o, use -DBOARD=RAD1O instead.
 
+If you get the "`arm-none-eabi-gcc` is not a full path and was not found in the PATH"
+error during `cmake ..`, install the package with `sudo apt-get install gcc-arm-none-eabi`.
```
martinling, line 35: Hi, sorry for not reviewing this sooner. Looks fine but we should avoid giving distro-specific advice on how to install dependencies. I suggest the following change:

```suggestion
error during `cmake ..`, install the
[ARM GNU toolchain](https://developer.arm.com/Tools%20and%20Software/GNU%20Toolchain).
This toolchain is pre-packaged on many platforms and distributions, e.g. on Debian/Ubuntu
you can run `sudo apt-get install gcc-arm-none-eabi` to install it.
```

martinling (changes requested):

martinling (approved):
Thanks!

PR #1106, opened 2022-06-13
Implement NXP's recommended setup sequence for the PLL and M4 clock

martinling:
This follows the sequence described in:

UM10503 Rev 2.4 (Aug 2018), section 13.2.1.1, page 167


mossmann:
I tested this with three different units that all have experienced intermittent start-up failures under certain conditions. This fixed all three of them!

PR #1112, opened 2022-06-27
Standardise and enforce code style

martinling:
This PR applies a [clang-format](https://clang.llvm.org/docs/ClangFormat.html) configuration that has been chosen to minimise change to the existing codebase, whilst still resulting in a consistent style across the project that can be applied and verified automatically. The configuration chosen does not conform to any specific human-readable style guide. It requires clang-format version 14 or higher.

First there are several commits which manually clean up code and comments which cannot be reformatted well automatically.

Next, `// clang-format off` is applied to areas with manual layout to be preserved.

Then the clang-format configuration is added, along with a new `tools/reformat-source.sh` script to apply it, and a formatting check is added to the CI.

Finally in the last commit, all code is reformatted to meet the new standard.


mossmann:
Can you please change this from Jenkins to GitHub Actions? I've temporarily disabled Jenkins for this repo, and even when it can be returned to service I would prefer to use Jenkins only for things that require hardware.

martinling:
Updated as follows:

- The CI check now uses [this GitHub Action](https://github.com/marketplace/actions/clang-format-check).
- Added a further commit which adds braces to any control statements that are missing them. This is not currently checked by `clang-format`, but we can add that check with the new `InsertBraces` option once version 15 is released.

We may still want to tweak the clang-format configuration further.

At the moment, the reformat touches about 15-20% of the code by line count. The biggest source of changes is the formatting of function types, with the rest mostly related to inconsistent whitespace. To some extent the function churn is inevitable since we were inconsistent before, but we should make sure we're happy with what we choose going forward.

What I have at the moment could be summarised as:

1. If the function prototype fits within 100 characters, put it on one line including the opening brace:
```c
return_type function_on_one_line(type_one arg_one, type_two arg_two) {
        ...
}
```

2. Otherwise, put each argument on its own line, with the closing bracket and opening brace on a new line:
```c
return_type function_on_multiple_lines(
        type_one long_argument_name_one,
        type_two long_argument_name_two,
        type_three long_argument_name_three
) {
        ...
}
```
The latter case doesn't actually match what we were doing before. Usually we put the closing bracket after the last argument, and the opening brace on its own line. This can be done by changing `AlignAfterOpenBracket` from `BlockIndent` to `AlwaysBreak`, and setting `BraceWrapping: AfterFunc` to `true`.

However, doing that creates two other problems:
1. Setting `BraceWrapping: AfterFunc` to `true` means that the opening brace will _always_ be on its own line, even when the function prototype is short. This creates both a lot of churn and wasted space - we _usually_ prefer a single line when possible, although notably not in libhackrf.

2. The `AlignAfterOpenBracket` setting is applied to function _calls_ as well, not just declarations. And when we split calls over multiple lines, we usually prefer to put the closing bracket on a new line - we do this a lot in `libusb_control_transfer` calls in libhackrf, for instance, and I like it that way, because it adds a visual end to the indented arguments.

The other option would be `AlignAfterOpenBracket: AlwaysBreak` and `BraceWrapping: AfterFunc: false`, but I don't like that combination because it leaves no visual separation between the last argument and the first line of a function:

```c
return_type function_on_multiple_lines(
        type_one long_argument_name_one,
        type_two long_argument_name_two,
        type_three long_argument_name_three) {
        first_line();
        second_line();
}
```
So for function protototypes it's basically a question of choosing which available combination to go with, unless we choose another linter entirely (I tried uncrustify, but found it a lot harder to get it close to our style overall).

Besides that, I'm mostly happy with the output, although there are some line wrapping decisions I dislike. These could probably be improved a bit by adjusting the various penalty factors in the configuration.

host/hackrf-tools/src/hackrf_info.c
```diff
@@ -26,126 +26,144 @@
...
+	hackrf_device_list_t* list;
 	hackrf_device* device;
 	int i, j;
 
 	result = hackrf_init();
 	if (result != HACKRF_SUCCESS) {
-		fprintf(stderr, "hackrf_init() failed: %s (%d)\n",
-				hackrf_error_name(result), result);
+		fprintf(stderr, "hackrf_init() failed: %s (%d)\n", hackrf_error_name(result), result
+		);
```
metayan, line 43: The automatic reformatting causes plenty of these hanging parens.

host/hackrf-tools/src/hackrf_sweep.c
```diff
@@ -349,22 +353,23 @@ static void usage() {
 	fprintf(stderr, "\t[-p antenna_enable] # Antenna port power, 1=Enable, 0=Disable\n");
 	fprintf(stderr, "\t[-l gain_db] # RX LNA (IF) gain, 0-40dB, 8dB steps\n");
 	fprintf(stderr, "\t[-g gain_db] # RX VGA (baseband) gain, 0-62dB, 2dB steps\n");
-	fprintf(stderr, "\t[-w bin_width] # FFT bin width (frequency resolution) in Hz, 2445-5000000\n");
+	fprintf(stderr,
```
metayan, line 356: Hmm.
Maybe these long printout blocks should be handled as one long string anyway.

host/hackrf-tools/src/hackrf_transfer.c
```diff
@@ -749,56 +754,72 @@ int main(int argc, char** argv) {
...
-			fprintf(stderr, "argument error: if_freq_hz must be specified for explicit tuning.\n");
+			fprintf(stderr,
+				"argument error: if_freq_hz must be specified for explicit tuning.\n"
+			);
 			usage();
 			return EXIT_FAILURE;
 		}
 		if (!image_reject) {
-			fprintf(stderr, "argument error: image_reject must be specified for explicit tuning.\n");
+			fprintf(stderr,
```
metayan, line 797: Might there be a setting to not reformat fprintf's at all?
Maybe until someone comes up with a better algorithm. ;)

host/libhackrf/src/hackrf.c
```diff
@@ -1310,9 +1250,7 @@ int ADDCALL hackrf_version_string_read(hackrf_device* device, char* version,
 	}
 }
 
-extern ADDAPI int ADDCALL hackrf_usb_api_version_read(hackrf_device* device,
-		uint16_t* version)
-{
+extern ADDAPI int ADDCALL hackrf_usb_api_version_read(hackrf_device* device, uint16_t* version) {
```
metayan, line 1253: Sometimes the automatic formatting goes against the manual efforts to increase readability. Which makes me wonder if it's really a good idea to hand over the intelligent efforts of the programmer to a rigid algorithm. It does well in many cases, and surely has some value, but at what cost? (Just a general point, because I really appreciate your efforts and see that a lot of the changes are increasing readability and understanding of the code. And I'm only criticizing the automatic reformatting. Maybe the settings could be tweaked further?)

host/libhackrf/src/hackrf.h
```diff
@@ -27,31 +27,31 @@ ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSI
 #include <stdint.h>
```
metayan, line 27: Maybe exclude this file from auto-formatting altogether.
It messes with defines and uglifies most function prototypes.

host/hackrf-tools/src/hackrf_sweep.c
martinling: Maybe, but there are several parameters interpolated into that one big string, and it's much easier to see where each ones fit in the format string when it's done line by line.

The best way to improve this section might be to add an `err()` macro rather than writing out `fprintf(stderr, ...)` every time.

host/hackrf-tools/src/hackrf_transfer.c
martinling: No, clang-format doesn't give that level of control. But see note above about using a macro/wrapper for these.

host/libhackrf/src/hackrf.c
martinling: I don't see any manual effort to improve readability here, just a line that was previously wrapped in 80 characters which has now been left on a single line because we chose a line width of 100 characters.

We could choose a shorter line width - I started with 100 because it seemed to largely reflect existing practice.

One argument for a shorter length would be that GitHub ends up displaying a horizontal scrollbar for many of the diffs you've commented on...

host/hackrf-tools/src/hackrf_info.c
martinling: Fixed by no longer using the `BlockIndent` setting.

host/hackrf-tools/src/hackrf_sweep.c
martinling: I've replaced this whole block with a single call to `fprintf`, using string literal concatenation for the lines.

host/libhackrf/src/hackrf.c
martinling: I've reduced the line width to 90 characters, so this function header is now split over multiple lines again.

metayan:
Nicely done.

I tried adding
`Cpp11BracedListStyle: false`
to get some space in {} lists but it does a bit more than I was hoping for.
A simple `SpacesInBraces` option matching the other `SpacesIn...` options would be nice. Some day. ;)

host/libhackrf/src/hackrf.h
```diff
@@ -27,31 +27,31 @@ ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSI
...
-  /* Define calling convention in one place, for convenience. */
-  #define ADDCALL __cdecl
+#define ADD_EXPORTS
+
+/* You should define ADD_EXPORTS *only* when building the DLL. */
+#ifdef ADD_EXPORTS
+#define ADDAPI __declspec(dllexport)
+#else
+#define ADDAPI __declspec(dllimport)
+#endif
```
martinling, line 37: There is an option to indent statements inside an `#ifdef` block if we want that - it wasn't something we were consistent about before so I left it off.

host/libhackrf/src/hackrf.h
```diff
@@ -182,88 +182,170 @@ typedef struct {
...
+	hackrf_device* device,
+	hackrf_m0_state* value);
+
+extern ADDAPI int ADDCALL hackrf_set_tx_underrun_limit(
+	hackrf_device* device,
+	uint32_t value);
+
+extern ADDAPI int ADDCALL hackrf_set_rx_overrun_limit(
+	hackrf_device* device,
+	uint32_t value);
```
martinling, line 250: The `SeparateDefinitionBlocks` option doesn't add spacing between function declarations. Since this file is directly user-facing, I've manually added empty lines between the declarations to make it easier to read. Having them bunched up next to each other works well when they're each on single lines, but less so when they're wrapped.

In any case these should all be gaining doxygen comments at some point.

martinling (commented):
Added some comments on parts of the diff that we may still want to tweak.

firmware/common/cpld_xc2c.c
```diff
@@ -401,13 +477,22 @@ bool cpld_xc2c64a_jtag_sram_verify(
...
+		const int data_row = (int) address_row - 1;
+		const size_t mask_index =
+			(data_row >= 0) ? verify->mask_index[data_row] : 0;
+		const uint8_t* const expected =
+			(data_row >= 0) ? &program->row[data_row].data[0] : NULL;
+		const uint8_t* const mask =
+			(data_row >= 0) ? &verify->mask[mask_index].value[0] : NULL;
+		const uint8_t next_address = (address_row < CPLD_XC2C64A_ROWS) ?
+			cpld_hackrf_row_addresses.address[address_row] :
+			0;
```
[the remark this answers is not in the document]
martinling: Oh, I guess it's because it split the ternary expression, and when it does so it always puts both results on their own lines. Which makes sense as a rule, but just looks weird here because the `0` is so much shorter than the other result.

host/libhackrf/src/hackrf.h
metayan: Even though the nesting of defines in hackrf isn't complex, I find the indentation made it very clear. (Was a bit surprised that it was removed.) Some points about indenting defines: https://stackoverflow.com/questions/789073/indenting-defines

martinling: I'm fine with adding it to the places we were missing it. Have gone ahead and done that - I had to manually help it indent some multiline comments within `#ifdef` blocks though.

host/libhackrf/src/hackrf.h
metayan: Noticed that one can stick a linefeed between blocks of `#define`s to prevent the weird columnizing.

martinling: The 'weird columnizing' is the `AlignConsecutiveMacros` option. It's something we were doing by hand a lot, so I turned it on, but of course that means it's now applied everywhere.

It has sub-options for whether to align `AcrossEmptyLines` and `AcrossComments`, which are both off by default, so you can always prevent it doing that by separating things.

For the most part I'm happy with it since we tend to group related definitions and separate unrelated ones anyway.

metayan: Wasn't trying to nitpick – just that the first thing that I happened to see was
```
#define SAMPLES_PER_BLOCK                8192
#define BYTES_PER_BLOCK                  16384
#define MAX_SWEEP_RANGES                 10
#define HACKRF_OPERACAKE_ADDRESS_INVALID 0xFF
#define HACKRF_OPERACAKE_MAX_BOARDS      8
#define HACKRF_OPERACAKE_MAX_DWELL_TIMES 16
#define HACKRF_OPERACAKE_MAX_FREQ_RANGES 8
```
which gave me an aesthetic jolt, but with a linefeed it became
```
#define SAMPLES_PER_BLOCK 8192
#define BYTES_PER_BLOCK   16384
#define MAX_SWEEP_RANGES  10

#define HACKRF_OPERACAKE_ADDRESS_INVALID 0xFF
#define HACKRF_OPERACAKE_MAX_BOARDS      8
#define HACKRF_OPERACAKE_MAX_DWELL_TIMES 16
#define HACKRF_OPERACAKE_MAX_FREQ_RANGES 8
```
which calmed me and instilled hope...
So wanted to inform you about the trick.

martinling: Yeah, definitely makes sense to separate those two groups. I'll add a commit that splits them up. Let me know if you spot any others.

host/hackrf-tools/src/hackrf_cpldjtag.c
```diff
@@ -31,23 +31,23 @@
...
 #ifdef _MSC_VER
-#ifdef _WIN64
+	#ifdef _WIN64
 typedef int64_t ssize_t;
-#else
+	#else
 typedef int32_t ssize_t;
-#endif
+	#endif
 #endif
```
martinling, line 39: This combination, of indenting the preprocessor statements but not the declarations between them, seems a bit odd.

host/libhackrf/src/hackrf.h
martinling: Done.

firmware/common/gpio_lpc.h
```diff
@@ -53,18 +53,20 @@ struct gpio_t {
...
-#define GPIO(_port_num, _pin_num) { \
-	.mask = (1UL << (_pin_num)), \
-	.port = GPIO_LPC_PORT(_port_num), \
-	.gpio_w = GPIO_LPC_W(_port_num, _pin_num), \
-}
+#define GPIO(_port_num, _pin_num)                                              \
+	{                                                                      \
+		.mask = (1UL << (_pin_num)), .port = GPIO_LPC_PORT(_port_num), \
+		.gpio_w = GPIO_LPC_W(_port_num, _pin_num),                     \
+	}
```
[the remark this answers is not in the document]
martinling: Done.

host/hackrf-tools/src/hackrf_cpldjtag.c
```diff
@@ -129,13 +129,14 @@ int main(int argc, char** argv)
...
-	{
+	if (infile == NULL) {
 		fprintf(stderr, "Failed to open file: %s\n", path);
 		return EXIT_FAILURE;
 	}
 	/* Get size of the file  */
-	fseek(infile, 0, SEEK_END); /* Not really portable but work on major OS Linux/Win32 */
+	fseek(infile,
+	      0,
+	      SEEK_END); /* Not really portable but work on major OS Linux/Win32 */
```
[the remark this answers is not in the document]
martinling: Done.

martinling:
Rebased after the recent PR marges. Also added a few more manual commits before the big reformat, to improve some odd output noticed on the previous review pass.

mossmann (approved):
Lovely! Thank you @martinling and @metayan!

PR #1116, opened 2022-07-07
Fix #902 - Report transmit amplitude

### LINKED ISSUE #902: `hackrf_transfer` reports bogus transmit amplitude
### Steps to reproduce
`hackrf_transfer -t /dev/zero` or `hackrf_transfer -c 127`

### Expected behaviour
Amplitude, if reported at all, should be computed from the samples sent over USB.

### Actual behaviour
```
19.9 MiB / 1.000 sec = 19.9 MiB/second, amplitude -inf dBfs
19.9 MiB / 1.000 sec = 19.9 MiB/second, amplitude -inf dBfs
19.9 MiB / 1.000 sec = 19.9 MiB/second, amplitude -inf dBfs
```

gitcolt:
`stream_amplitude` was only being updated in `rx_callback()` and not in `tx_callback()`.

### DIFF
--- a/host/hackrf-tools/src/hackrf_transfer.c
+++ b/host/hackrf-tools/src/hackrf_transfer.c
@@ -1149,7 +1149,7 @@ int main(int argc, char** argv) {
 			    // This is only an approximate measure, to assist getting receive levels right:
 			    double	full_scale_ratio = ((double)stream_amplitude_now / (byte_count_now ? byte_count_now : 1))/128;
 			    double	dB_full_scale_ratio = 10*log10(full_scale_ratio);
-					if (dB_full_scale_ratio > 1 || dB_full_scale_ratio == -INFINITY) // Guard against ridiculous reports
+					if (dB_full_scale_ratio > 1) // Guard against ridiculous reports
 						dB_full_scale_ratio = -0.0;
 			    fprintf(stderr, "%4.1f MiB / %5.3f sec = %4.1f MiB/second, amplitude %3.1f dBfs",
 				    (byte_count_now / 1e6f),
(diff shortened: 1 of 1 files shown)


host/hackrf-tools/src/hackrf_transfer.c
```diff
@@ -1146,8 +1149,8 @@ int main(int argc, char** argv) {
 			    // This is only an approximate measure, to assist getting receive levels right:
 			    double	full_scale_ratio = ((double)stream_amplitude_now / (byte_count_now ? byte_count_now : 1))/128;
 			    double	dB_full_scale_ratio = 10*log10(full_scale_ratio);
-			    if (dB_full_scale_ratio > 1)
-			    	dB_full_scale_ratio = NAN;	// Guard against ridiculous reports
+					if (dB_full_scale_ratio > 1 || dB_full_scale_ratio == -INFINITY) // Guard against ridiculous reports
+						dB_full_scale_ratio = -0.0;
```
mossmann, line 1153: Is there a reason you added the `dB_full_scale_ratio == -INFINITY` condition that is still relevant now that your other changes are applied? There is a fairly common test case of `hackrf_transfer -t /dev/zero` which this now reports as full scale instead of "-inf dBfs" which I think is more appropriate.

I like the change to `-0.0`, but I would like to see the `-INFINITY` condition removed.

mossmann (changes requested):
Please make the requested change.

host/hackrf-tools/src/hackrf_transfer.c
gitcolt: My HackRF One reports -inf when reading from /dev/zero. Am I understanding correctly that this is the intended behavior?
image

I made the suggested change in my last commit.

mossmann: Yes, that is the intended behavior. Decibels are weird in this case because they describe a ratio where there is a zero in the numerator. Maximum power is 0 dBfs. Half power is -3 dBfs. Quarter power is -6 dBfs. Zero power is -infinity dBfs.

mossmann (approved):
Wonderful! Thank you very much!

PR #1131, opened 2022-08-15
Additional argument validation for hackrf_transfer

### LINKED ISSUE #1041: hackrf_transfer should check input parameters against accepted ranges.
Currently, `hackrf_transfer -s` doesn't check the sample rate or return an error if it is out of range.

This has led to some user confusion e.g: 

* Setting `-s 2` for 2MHz without realising that the number is in Hz not Mhz.
* Setting sample rates above 20MHz, which HackRF will try to use but fail to, resulting in uncontrolled behaviour.

The accepted ranges for all parameters are given in the usage message. We should check all of the parameters against these ranges.

martinling:
This PR adds some additional checks to arguments passed to `hackrf_transfer`:

- The `-s` option now checks that the sample rate is in the supported range of 2 to 20 MHz.
- The `-f` option now checks that the RF frequency is in the supported range of 1 to 6000 MHz.

- The new `-F` argument allows setting values outside the supported ranges above. The maximum RF frequency is still limited to 7250 MHz with this option, but no other limits are applied.

Closes #1041.

### DIFF
--- a/host/hackrf-tools/src/hackrf_transfer.c
+++ b/host/hackrf-tools/src/hackrf_transfer.c
@@ -379,6 +379,8 @@ uint32_t antenna_enable;
 bool sample_rate = false;
 uint32_t sample_rate_hz;
 
+bool force_sample_rate = false;
+
 bool limit_num_samples = false;
 uint64_t samples_to_xfer = 0;
 size_t bytes_to_xfer = 0;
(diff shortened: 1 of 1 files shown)


mossmann (approved):
Looks great! I made a minor change in one line to avoid a compiler warning and will merge after CI runs again.

PR #1132, opened 2022-08-15
Support repeating short files in `hackrf_transfer`

### LINKED ISSUE #720: hackrf_transfer -R (repeat) doesn't work with files smaller than the USB transfer buffer
ref #714

┆Issue is synchronized with this [Basecamp todo](https://3.basecamp.com/5295633/buckets/26082705/todos/4597433141) by [Unito](https://www.unito.io)

martinling:
This PR fixes repeated transmission of short sample files using `hackrf_transfer`:

- The first three commits simplify the TX and RX callbacks, by using early returns rather than deeply indented conditional blocks. There should be no functional change from these commits.

- The main change is in https://github.com/greatscottgadgets/hackrf/pull/1132/commits/0c35cff05b061afae12b804fffb37e51bfbc9835, which adds a loop which repeatedly rewinds the file until the transfer buffer is filled.

- Finally the message `Input file end reached. Rewind to beginning` is removed, because with short input files it can be printed an extremely large number of times.

Closes #720.

### DIFF
--- a/host/hackrf-tools/src/hackrf_transfer.c
+++ b/host/hackrf-tools/src/hackrf_transfer.c
@@ -414,67 +414,65 @@ int rx_callback(hackrf_transfer* transfer)
 	size_t bytes_written;
 	unsigned int i;
 
-	if (file != NULL) {
-		byte_count += transfer->valid_length;
-		bytes_to_write = transfer->valid_length;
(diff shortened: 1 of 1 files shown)


mossmann (approved):
lovely!

PR #1133, opened 2022-08-15
Overhaul timing in hackrf_transfer

### LINKED ISSUE #1019: Faster and cleaner exit from `hackrf_transfer`

martinling:
Rather than using `sleep` for 1s at a time, set up an interval timer that will fire once per second, and wait in the main loop for either this or some other event.

On POSIX, the timing is set up with `setitimer`, which generates a `SIGALRM` signal each time the timer fires. The main loop runs `pause` to wait for any signal.

On Windows, the timing is set up using `CreateWaitableTimer`, which provides an event handle that is set each time the timer fires. The main loop runs `WaitForMultipleObjects` to wait on this and an interrupt event.

The TX and RX callbacks can now stop the main loop immediately when they stop streaming. This fixes #1019.

### DIFF
--- a/host/hackrf-tools/src/hackrf_transfer.c
+++ b/host/hackrf-tools/src/hackrf_transfer.c
@@ -106,10 +106,6 @@ int gettimeofday(struct timeval* tv, void* ignored)
 #define BASEBAND_FILTER_BW_MIN (1750000)  /* 1.75 MHz min value */
 #define BASEBAND_FILTER_BW_MAX (28000000) /* 28 MHz max value */
 
-#if defined _WIN32
-	#define sleep(a) Sleep((a * 1000))
-#endif
-
 typedef enum {
 	TRANSCEIVER_MODE_OFF = 0,
 	TRANSCEIVER_MODE_RX = 1,
@@ -323,6 +319,9 @@ char* u64toa(uint64_t val, t_u64toa* str)
 }
 
 static volatile bool do_exit = false;
+#ifdef _WIN32
+static HANDLE interrupt_handle;
+#endif
 
 FILE* file = NULL;
 volatile uint32_t byte_count = 0;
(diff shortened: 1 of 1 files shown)


martinling:
Note that these changes could also be applied to `hackrf_sweep`, but I figure it makes sense to review one copy of them first.

mossmann:
This seems like a nice change, but I don't think it entirely implements #1019. In particular, if `-n` is used to transfer a number of samples that takes less than one second, this still executes for a full second.

martinling:
Oh, hmm - how are you testing to see that result?

The early return in the limit case is implemented [here](https://github.com/greatscottgadgets/hackrf/blob/a09e9a20ed42567b1fe979ab4db12f9f2590b2a4/host/hackrf-tools/src/hackrf_transfer.c#L463-L464) and [here](https://github.com/greatscottgadgets/hackrf/blob/a09e9a20ed42567b1fe979ab4db12f9f2590b2a4/host/hackrf-tools/src/hackrf_transfer.c#L538-L539). For me I see the `-n` option working as I'd expect:
```
$ hackrf_transfer -t /dev/zero -n 1000000
[...]
samples_to_xfer 1000000/1Mio
Stop with Ctrl-C
 2.1 MiB / 0.105 sec = 20.0 MiB/second, amplitude -inf dBfs
Exiting...
Total time: 0.10500 s
[...]

$ hackrf_transfer -r /dev/null -n 1000000
[...]
samples_to_xfer 1000000/1Mio
Stop with Ctrl-C
 2.1 MiB / 0.106 sec = 19.9 MiB/second, amplitude -20.4 dBfs
Exiting...
Total time: 0.10565 s
```

mossmann:
That's what I tried, but I must have made some error because I can't reproduce my earlier results.

mossmann (approved):
I love it!

PR #1135, opened 2022-08-17
Add a simpler way to check CLKIN status.

martinling:
Add a new `hackrf_clock` option (`-i` or `--clkin`) which reports the CLKIN status, and doesn't require remembering cryptic Si5351 register values.


mossmann (approved):
Excellent, thank you!

PR #1138, opened 2022-08-23
Preload TX transfers

martinling:
Currently when starting TX, libhackrf will pre-submit 4 x 256KB transfers of zeroes before any user-supplied samples. This adds a significant delay to transmission.

This PR makes `prepare_transfers` call the TX callback function to pre-fill those initial transfers with sample data. This means that the TX callback will immediately be called four times when `hackrf_start_tx` is called.

A similar change was included in #662, but the implementation here is somewhat different:

- At the time that PR was written, libhackrf started a new transfer thread for each TX or RX session. To implement preload, the `prepare_transfers` call for TX was moved into the transfer thread, before the libusb event handling loop. Now, libhackrf keeps a single transfer thread for each open HackRF device. In this implementation, the preload calls are made directly from the thread calling `hackrf_start_tx`.

- The TX callback is called to fill all four transfers before any are submitted. This ensures that we can't have a situation where a transfer completes and is refilled by the transfer thread before the preload step is completed. This prevents the TX callback from being called concurrently from two different threads, or buffers being filled in the wrong order.

### DIFF
--- a/host/libhackrf/src/hackrf.c
+++ b/host/libhackrf/src/hackrf.c
@@ -104,6 +104,9 @@ typedef enum {
 
 #define USB_CONFIG_STANDARD 0x1
 
+#define RX_ENDPOINT_ADDRESS (LIBUSB_ENDPOINT_IN | 1)
+#define TX_ENDPOINT_ADDRESS (LIBUSB_ENDPOINT_OUT | 2)
+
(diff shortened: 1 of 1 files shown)


mossmann (approved):
Very nice! Thank you!

PR #1139, opened 2022-08-24
Support flushing the device TX buffer before ending transmission

martinling:
This PR adds two new functions to libhackrf:

- `hackf_enable_tx_flush(device, enable)`: When called before `hackrf_start_tx`, this will make libhackrf submit an additional USB transfer containing 32KB of zeros after the last samples from the TX callback. This is the size of the device's TX buffer.

- `hackrf_await_tx_flush(device)`: When called after `hackrf_enable_tx_flush` this will wait until that additional USB transfer completes. Once that happens, it is guaranteed that all user samples have been transmitted, so it is safe to call `hackrf_stop_tx` without truncating the transmission.

These calls are then added to `hackrf_transfer`.

### DIFF
--- a/host/hackrf-tools/src/hackrf_transfer.c
+++ b/host/hackrf-tools/src/hackrf_transfer.c
@@ -1391,7 +1391,7 @@ int main(int argc, char** argv)
 	interval_timer.it_value.tv_sec = 0;
 	setitimer(ITIMER_REAL, &interval_timer, NULL);
 #endif
-	if (transmit && !interrupted) {
+	if ((transmit || signalsource) && !interrupted) {
 		// Wait for TX to finish.
 		hackrf_await_tx_flush(device);
 	}
--- a/host/hackrf-tools/src/hackrf_transfer.c
+++ b/host/hackrf-tools/src/hackrf_transfer.c
@@ -322,6 +322,7 @@ char* u64toa(uint64_t val, t_u64toa* str)
 }
 
 static volatile bool do_exit = false;
+static volatile bool interrupted = false;
 #ifdef _WIN32
 static HANDLE interrupt_handle;
 #endif
@@ -638,6 +639,7 @@ static hackrf_device* device = NULL;
 BOOL WINAPI sighandler(int signum)
 {
 	if (CTRL_C_EVENT == signum) {
+		interrupted = true;
 		fprintf(stderr, "Caught signal %d\n", signum);
 		stop_main_loop();
 		return TRUE;
@@ -647,6 +649,7 @@ BOOL WINAPI sighandler(int signum)
 #else
 void sigint_callback_handler(int signum)
 {
+	interrupted = true;
 	fprintf(stderr, "Caught signal %d\n", signum);
 	do_exit = true;
 }
@@ -1388,7 +1391,7 @@ int main(int argc, char** argv)
 	interval_timer.it_value.tv_sec = 0;
 	setitimer(ITIMER_REAL, &interval_timer, NULL);
 #endif
-	if (transmit) {
+	if (transmit && !interrupted) {
 		// Wait for TX to finish.
 		hackrf_await_tx_flush(device);
 	}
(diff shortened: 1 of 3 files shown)


mossmann:
Testing `hackrf_transfer` I do see the 32 KB of zeros, but I am unable to terminate the program with Ctrl-C (or signals such as SIGTGERM) when transmitting data from a file. It gets stuck in `hackrf_await_tx_flush()`. If transmitting a file without `-R`, the flush does eventually happen when the end of the file is reached or the number of samples specified with -n is exhausted. With `-R` it runs forever until I SIGKILL it.

host/libhackrf/src/hackrf.c
```diff
@@ -269,6 +275,9 @@ static int free_transfers(hackrf_device* device)
 		free(device->transfers);
 		device->transfers = NULL;
 	}
+
+	libusb_free_transfer(device->flush_transfer);
```
mossmann, line 279: Should this be conditional to avoid a double free?

host/libhackrf/src/hackrf.c
```diff
@@ -1898,6 +1936,49 @@ int ADDCALL hackrf_start_tx(
...
+			TX_ENDPOINT_ADDRESS,
+			calloc(1, DEVICE_BUFFER_SIZE),
+			DEVICE_BUFFER_SIZE,
+			hackrf_libusb_flush_callback,
+			device,
+			0);
+
+		device->flush_transfer->flags = LIBUSB_TRANSFER_FREE_BUFFER;
+	} else {
+		libusb_free_transfer(device->flush_transfer);
```
mossmann, line 1962: Should this be conditional to avoid a double free?

host/hackrf-tools/src/hackrf_transfer.c
```diff
@@ -1387,6 +1388,10 @@ int main(int argc, char** argv)
 	interval_timer.it_value.tv_sec = 0;
 	setitimer(ITIMER_REAL, &interval_timer, NULL);
 #endif
+	if (transmit) {
+		// Wait for TX to finish.
+		hackrf_await_tx_flush(device);
```
mossmann, line 1393: This can wait forever (see my previous PR comment) because `hackrf_stop_tx()` hasn't been called yet. I think the right place for this is where "hackrf_stop_tx() done\n" is printed. I would like flushing to happen in `signalsource` mode in addtion to `transmit` mode (for the case where `-n` is used with `-c`).

mossmann (changes requested):

host/libhackrf/src/hackrf.c
martinling: It doesn't need to be, the function is [documented](https://libusb.sourceforge.io/api-1.0/group__libusb__asyncio.html#ga6ab8b2cff4de9091298a06b2f4b86cd6) to be a no-op with a null pointer.

host/hackrf-tools/src/hackrf_transfer.c
martinling: That `hackrf_stop_tx()` hasn't been called yet is exactly the intent. We want to delay that call until the buffer is flushed, because once we make that call the device stops TX immediately.

The bug you found is that I didn't handle the case where the main loop is interrupted by Ctrl-C, rather than by the TX callback calling `stop_main_loop` before returning nonzero. In that scenario we shouldn't wait for a flush at all - the user wants to stop, so we should just call `hackrf_stop_tx` immediately.

host/libhackrf/src/hackrf.c
martinling: (also note that I set the `LIBUSB_TRANSFER_FREE_BUFFER` flag on that transfer, so this call will free the 32KB buffer as well)

host/hackrf-tools/src/hackrf_transfer.c
martinling: Fixed in new commits.

mossmann (approved):
Excellent! Thanks for the explanations. Ctrl-C works now. I see flushes even when I use Ctrl-C, but I think that is fine. (Flushes may be generated, but we do not wait for them.)

martinling:
Yes, it's expected that the flushes will still be generated - we have to make the call to tell libhackrf that we're going to want to flush before we call `hackrf_start_tx`.

martinling:
(I guess we could omit the call to `hackf_enable_tx_flush` in cases where Ctrl-C is the only end condition, like `-R` or `-c` without `-n`, but I don't see any reason to)

PR #1149, opened 2022-09-01
operacake: disable the usage of time mode when a portapack is installed

### LINKED ISSUE #1137: Opera Cake could interfere with PortaPack in time mode
When an Opera Cake and a PortaPack are installed on the same HackRF One, the Opera Cake should be able to function in manual mode or frequency mode, but time mode is not supported. An attempt to use time mode (or other GPIO/SCT operation) would interfere with the PortaPack, so the firmware should refuse to enter time mode when a PortaPack is present.

grvvy:
Fixes #1137


mossmann (approved):
Perfect! Thank you!

PR #1150, opened 2022-09-02
Non-root android support

AlexandreRouma:
This PR contains the following modifications:

1) Fixed the cmake to allow compilation with the Android NDK
2) Disabled libusb enumeration when compiling on android since it is not supported (As explained [in the libusb documentation](https://github.com/libusb/libusb/blob/master/android/README) line 92)
3) Add the `hackrf_open_by_fd` function that allows to open a hackrf device using a file descriptor. This is the only possible way to open a device on an non-rooted android device.

This code has been tested and confirmed working on my serveral android device but I still recommend testing it further.
It should not impact any other operating system.
image


martinling:
Hi @AlexandreRouma, thanks for the PR!

The changes look straightforward, but as you may have seen there's a build failure on the CI - this is on Appveyor, which runs a Windows build. Since Windows doesn't provide an fd for USB devices I suggest we just wrap `hackrf_open_by_fd` with `#ifndef _WIN32`.

My other thought is about the use of `LIBUSB_OPTION_NO_DEVICE_DISCOVERY` due to this not currently working on Android. There's a libusb branch aiming to fix that problem which has been getting a lot of traction: https://github.com/libusb/libusb/pull/874, later updated at https://github.com/libusb/libusb/pull/1164. It would be good to avoid making changes in libhackrf that would prevent discovery from working if the libusb issue is fixed. Can we make this conditional somehow?

Finally, what is the Android app in your screenshot? Could you point us to some instructions for building this with libhackrf so that we can test it?

AlexandreRouma:
Hi,

I'll add the windows `ifdef` in a sec, didn't realize that libusb completely removes the function instead of simply returning LIBUSB_ERROR_NOSUPPORTED like basically all other functions. I should bring this up to the libusb team for more consistency.

The branches that aim to fix the enumeration support do exists but I have very high doubts that they will ever work. Still, it can indeed be made conditional. 

The app in the screenshot is an android build of my [SDR++](https://github.com/AlexandreRouma/SDRPlusPlus) project. It is currently very hard to build for android and still a bit buggy (not the fault of libhackrf). So I don't think it's an ideal way to test libhackrf on android.

AlexandreRouma:
not exactly sure why the CI's complaining about code style when it perfectly matches the rest of the file

straithe:
Did you start your work on this PR from the latest version in the repository?

AlexandreRouma:
yes, it was pulled yesterday a few hours before the PR

straithe:
Thank you for the update. I'll check this again next week when I'm back in the office (if no one else gets to you before then).

martinling:
We have a `clang-format` config in the repository which is checked on the CI. If you have `clang-format` installed, you can run `./tools/reformat-source.sh` to make the code match.

martinling:
Do you have any recommended process for building and testing libhackrf on android? I've done a little work with the NDK and Android dev tools in the past but it was some years ago.

AlexandreRouma:
Building for the NDK is quite involved because the dependencies must be compiles as well, so, so I'd suggest following the scripts I wrote for my [android-sdr-kit](https://github.com/AlexandreRouma/android-sdr-kit) project that builds all of these libraries for android (or just run the container).

Then, to test, you need to write an NDK app and link as usual. All you have to do is, in the java part, enumerate devices and locate the hackrf using its vid/pid combo, then request permission, and when you've got permission, send the file descriptor android gives you. After using the new `hackrf_open_by_fd` function, you can use everything as normal.

AlexandreRouma:
Hi, 
Is there any update on this?

mossmann:
> Hi, Is there any update on this?

We've been very busy preparing for a release for Opera Cake support which will be out very soon. We will give this a proper review after the release and hope to incorporate it into the next release. Thank you for your patience!

host/libhackrf/src/hackrf.c
```diff
@@ -754,6 +759,36 @@ int ADDCALL hackrf_open_by_serial(
 	return hackrf_open_setup(usb_device, device);
 }
 
+int ADDCALL hackrf_open_by_fd(
+	int fd,
```
miek, line 763: I think we should take `intptr_t sys_dev` here instead and pass it straight through to libusb. Looking back at at the PR for the libusb function (https://github.com/libusb/libusb/pull/242#issuecomment-444600509), they made that choice to make it more general across different platforms so I think we should do the same and leave it up to the user to do any platform-specific conversion/checking.

host/libhackrf/CMakeLists.txt
```diff
@@ -33,6 +33,11 @@ include(${PROJECT_SOURCE_DIR}/../cmake/set_release.cmake)
 add_definitions(-DLIBRARY_RELEASE="${RELEASE}")
 set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/../cmake/modules)
 
+option(DISABLE_USB_ENUMERATION "Prevent libusb from trying to enumerate devices. Useful on non-root android" OFF)
```
miek, line 36: I'd rather name this something like `DISABLE_USB_DEVICE_DISCOVERY`. I think it's clearer if it matches libusb's name & doesn't get confused with OS-level enumeration.

host/libhackrf/src/hackrf.c
```diff
@@ -754,6 +759,36 @@ int ADDCALL hackrf_open_by_serial(
...
+
+	if (fd < 0) {
+		return HACKRF_ERROR_INVALID_PARAM;
+	}
+
+	if (device == NULL) {
+		return HACKRF_ERROR_INVALID_PARAM;
+	}
+
+	int err = libusb_wrap_sys_device(g_libusb_context, (intptr_t)fd, &usb_device);
```
miek, line 777: This function was added fairly recently (in libusb 1.0.23) so I think we should add an ifdef on `LIBUSB_API_VERSION` and return unsupported otherwise, to allow for building with older versions.

straithe:
@AlexandreRouma I'm checking in to see if you've seen the requested changes. Please let me know!

AlexandreRouma:
Hi,

Sorry, had some stuff come up IRL, didn't have as much time for coding and completely forgot about the PR.
Will try to implement this weekend if I get the time.

straithe:
Hope you are doing ok! I'll watch out for an update.

CRD716:
@AlexandreRouma Any plans to complete this?

martinling:
Some of the CMake changes required here are included in #1586.

PR #1153, opened 2022-09-09
Measure digital signal power, not amplitude in hackrf_transfer

mossmann:
This is a rewrite of the amplitude measurement and reporting that greatly improves accuracy and reliability.

One remaining potential source of unreliability is the fact that `byte_count` and `stream_power` are not written, read, or reset atomically as a pair, but making those operations adjacent so that they happen at approximately the same time eliminated all symptoms of concurrency problems in tests on my system. I think that locking (or a lock-free ring buffer for stats) would be required to guarantee 100% reliability, but I don't feel that it is worth that complication. The worst thing that could happen is that the reported stats are a bit off.

Additionally this updates the `-c` option to accept CW amplitude configuration up to 128. Previously it was limited to 127 (maximum `uint8_t`), but it now uses negative sample values internally to support 128 as well. This makes it possible use `-c` to transmit at true absolute maximum power and is helpful for testing and demonstrating the new power measurement. (`-c 128` is exactly 0 dBfs, `-c 64` is -6.0 dBfs, etc.)

This also improves reporting at the start of a transmission by skipping any stats that are collected from empty buffers when `tx_callback()` is called to preload USB transfer buffers.

### DIFF
--- a/host/libhackrf/src/hackrf.h
+++ b/host/libhackrf/src/hackrf.h
@@ -25,6 +25,7 @@ ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSI
 #define __HACKRF_H__
 
 #include <stdint.h>
+#include <sys/types.h>
 
 #ifdef _WIN32
 	#define ADD_EXPORTS
@@ -457,6 +458,10 @@ extern ADDAPI int ADDCALL hackrf_start_rx_sweep(
 	hackrf_device* device,
 	hackrf_sample_block_cb_fn callback,
 	void* rx_ctx);
+
+extern ADDAPI size_t ADDCALL hackrf_get_transfer_buffer_size(hackrf_device* device);
+
+extern ADDAPI uint32_t ADDCALL hackrf_get_transfer_queue_depth(hackrf_device* device);
 
 #ifdef __cplusplus
 } // __cplusplus defined.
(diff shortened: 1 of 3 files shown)


martinling (commented):
This looks good, I've done some quick tests and it seems to work as I would expect.

One thought I have is that it might be better to add an API function to get the transfer count and transfer size, rather than exposing them in the header. There's been discussion of making these configurable, as it would help with some use cases that need to minimise latency. Requiring an API call to get the current settings would ensure forward compatibility if we make them configurable later. I think they should be configurable on a per-device basis.

martinling:
Yes, that sounds reasonable. Or perhaps something like `hackrf_get_transfer_queue_depth()` rather than `hackrf_get_transfer_count()`? The latter sounds more like something you'd use to retrieve a current count (which may also be something we want, in future).

host/libhackrf/src/hackrf.h
```diff
@@ -458,6 +458,10 @@ extern ADDAPI int ADDCALL hackrf_start_rx_sweep(
 	hackrf_sample_block_cb_fn callback,
 	void* rx_ctx);
 
+extern ADDAPI size_t ADDCALL hackrf_get_transfer_buffer_size(void);
+
+extern ADDAPI uint32_t ADDCALL hackrf_get_transfer_queue_depth(void);
+
```
martinling, line 464: Should we make these take a `hackrf_device*`? There may be scenarios where someone's using e.g. one HackRF for RX and one for TX, with different latency/throughput considerations.

martinling (approved):
Looks good to me!

PR #1154, opened 2022-09-10
firmware: use consistent clock source

mossmann:
CLKOUT now switches to the active clock source whenever the internal clocks change source. It previously did so only when (re)enabled.


martinling (approved):
Much tidier and less surprising!

PR #1155, opened 2022-09-14
Detect hardware platform at run-time.

mossmann:
Firmware now detects the hardware it is running on at startup and refuses to run if it is compiled for the wrong platform. The board ID returned by firmware to the host is now derived from run-time detection rather than a compile-time value. A separate method to retrieve compile-time supported platform is added.

On HackRF One, pin straps are checked to determine hardware revision. This is informational to aid troubleshooting and does not affect any function.

### DIFF
--- a/firmware/common/hackrf_core.h
+++ b/firmware/common/hackrf_core.h
@@ -42,23 +42,6 @@ extern "C" {
 #include "rf_path.h"
 #include "cpld_jtag.h"
 
-/* hardware identification number */
-#define BOARD_ID_JAWBREAKER 1
-#define BOARD_ID_HACKRF_ONE 2
-#define BOARD_ID_RAD1O      3
-
-#ifdef JAWBREAKER
-	#define BOARD_ID BOARD_ID_JAWBREAKER
-#endif
-
-#ifdef HACKRF_ONE
-	#define BOARD_ID BOARD_ID_HACKRF_ONE
-#endif
-
-#ifdef RAD1O
-	#define BOARD_ID BOARD_ID_RAD1O
-#endif
-
 /*
  * SCU PinMux
  */
--- a/firmware/common/platform_detect.c
+++ b/firmware/common/platform_detect.c
@@ -1,0 +1,231 @@
+/*
+ * Copyright 2022 Great Scott Gadgets
+ *
+ * This file is part of HackRF.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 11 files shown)


martinling (approved):
Looks good to me. I've tested with two r4 boards and a rad1o with old and new firmware. I haven't checked anything about the correctness for other hardware.

straithe:
I am confused. Can you please open a new issue with more detail about what you are asking?

mossmann:
It's a message among developers, and it has enough detail. :)
@martinling and I were just discussing this earlier and how unlikely it would be to affect anyone, but we were thinking only of BOARD_ID_HACKRF_ONE and hadn't considered BOARD_ID_INVALID! I'll take a look.

martinling:
It seems Soapy is only using `BOARD_ID_INVALID` as a placeholder to set when it hasn't run `hackrf_board_id_read` yet, to avoid leaving a variable uninitialized in the case where that call fails.

Even if this gets changed in Soapy now, it would be bad if we were to break building older Soapy versions with newer libhackrf.

So I think our options are to either rename `BOARD_ID_UNDETECTED` back to `BOARD_ID_INVALID`, or `#define` the latter to the former.

willcode:
The only place I've found was in soapyhackrf, where it's just used unnecessarily as an initializer, messing up the compile. My recommendation for them is just to init to 0. They're just calling the string version immediately with the result, so they don't actually care what the value is.

gr-osmosdr doesn't use this, so it may not be much of a practical issue.

mossmann:
I agree that BOARD_ID_UNDETECTED is appropriate for the initializations in https://github.com/pothosware/SoapyHackRF/pull/36.

We do version libhackrf, but we only take care to bump it once per release. We do not guarantee API compatibility across all git commits between releases. It has been bumped from 0.6 to 0.7 since the last release.

I like the idea of restoring `BOARD_ID_INVALID` (and `BOARD_ID_HACKRF_ONE` as well) with a `#define`.

I don't care for the idea of using `0` which represents a specific no-longer-supported hardware platform.

mossmann:
Another option would be to restore `BOARD_ID_INVALID` and give `BOARD_ID_UNDETECTED` (which is more meaningful in the firmware) a different number, but I propose we `#define` both the outdated IDs to make this a nonissue for any user of the API.

willcode:
Ok, I see where you get the version string dynamically in `hackrf_version_string_read()`. I'd only looked for a static value in the include. Yah, `#define`s will certainly fix this for SoapyHackRF.

PR #1156, opened 2022-09-14
Fix handling of the final <256KiB of data transmitted with `hackrf_transfer`

martinling:
This PR fixes handling of the final samples in a transmission sent with `hackrf_transfer`.

Previously, `hackrf_transfer` would drop the last < 256KiB from of any source of samples, when it could not populate a full transfer buffer.

The changes are:

- Some general cleanup in the `hackrf_transfer` TX callback, to clarify things and avoid duplicated code.
- Make libhackrf honour the `valid_length` set by a TX callback and put this in the libusb transfer length.
- In the `hackrf_transfer` TX callback, set `valid_length` to the bytes put in the buffer when filling a final short transfer.
- After filling the final transfer, return 0 from the callback to submit it, postponing the -1 return until the next call.

### DIFF
--- a/host/hackrf-tools/src/hackrf_transfer.c
+++ b/host/hackrf-tools/src/hackrf_transfer.c
@@ -507,16 +507,11 @@ int tx_callback(hackrf_transfer* transfer)
 		/* Transmit continuous wave with specific amplitude */
 		for (i = 0; i < bytes_to_read; i++)
 			transfer->buffer[i] = -(uint8_t) amplitude;
-
-		if (limit_num_samples && (bytes_to_xfer == 0)) {
-			stop_main_loop();
-			return -1;
-		} else {
-			return 0;
-		}
+		bytes_read = bytes_to_read;
+	} else {
+		bytes_read = fread(transfer->buffer, 1, bytes_to_read, file);
 	}
 
-	bytes_read = fread(transfer->buffer, 1, bytes_to_read, file);
 	if (limit_num_samples && (bytes_to_xfer == 0)) {
 		stop_main_loop();
 		return -1;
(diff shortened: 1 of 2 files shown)


martinling:
Further changes to correct some glitches seen in testing:

- Moved the `hackrf_start_tx` call below other setup calls in `hackrf_transfer`. This fixes samples being lost at the start due to the RF path not being fully configured yet.
- Submit the flush transfer introduced by #1139 earlier, as soon as the end of input data is reached. This avoids a gap which can otherwise occur in which the firmware is waiting for its own 16KB transfer request to complete before making the last samples available to the M0.
- Pad the last < 256KiB transfer with zeros to the next 512 byte boundary, to prevent the host sending a short packet with the last samples. This avoids a gap presumed to be due to USB scheduling details.

With these changes it's now possible to reliably go all the way down to `-n 1`.

RF output with `hackrf_transfer -c 127 -f 14e6 -x 30 -n 1`:

image

mossmann:
This hangs for me when I try to exit with Ctrl-C.

mossmann:
With Ctrl-C, it gets stuck in pthread_cond_wait() in cancel_transfers() in hackrf_stop_tx(). I've also now observed it getting stuck in the same place when using `-H 1` and trying to break out of the "Waiting for sync" state with Ctrl-C which seems to have happened ever since 8f60e6b.

martinling:
Thanks, have fixed that - we needed to disable flushing when cancelling transfers, or `transfer_finished` would skip decrementing `active_transfers` to zero because it thought there was still a flush transfer to complete.

mossmann (approved):
Nice! Thank you!

PR #1157, opened 2022-09-14
Standardise dBFS level

martinling:
[Decibels relative to full scale (dBFS)](https://en.wikipedia.org/wiki/DBFS) is defined in AES Standard AES17-1998, IEC 61606, and ITU-T Recs. P.381 and P.382, such that the RMS value of a full-scale sine wave is designated 0 dBFS.
    
The largest possible zero-centred sine wave on HackRF has the range -127 to 127. This PR corrects the calculation of dBFS in `hackrf_transfer` such that a sine wave of that amplitude should read 0 dBFS.

The `hackrf_transfer -c` option previously applied a constant amplitude to both I and Q, with the maximum of 128 creating a vector magnitude of ~181 and a level of +3 dBFS, previously reported as 0 dBFS. Instead, this PR applies the requested amplitude to the I channel only, setting Q to zero.

The maximum amplitude is limited to 127, such that the maximum supported level corresponds to 0 dBFS as defined above. Potentially, values of up to 181 could be accepted and converted to a vector of corresponding magnitude, but this is not implemented.

### DIFF
--- a/host/hackrf-tools/src/hackrf_transfer.c
+++ b/host/hackrf-tools/src/hackrf_transfer.c
@@ -1357,8 +1357,8 @@ int main(int argc, char** argv)
 				fprintf(stderr, "Waiting for sync...\n");
 			} else {
 				double full_scale_ratio = (double) stream_power_now /
-					(byte_count_now * 128 * 128);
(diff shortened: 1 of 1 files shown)


mossmann (approved):
Very nice!

PR #1161, opened 2022-09-18
clean-up and small bug fixes

mossmann:
Mostly this is just some simplification and clean-up. Fixes termination of `hackrf_transfer` when called from a script. Also fixes RX stats with small `-n`.


martinling (approved):
All looks good to me.

PR #1162, opened 2022-09-19
Fix antenna bias pulse glitch at startup.

### LINKED ISSUE #1006: Antenna port power turns on unexpectedly during HackRF power-on
When powering on a HackRF One, the antenna port power turns on briefly without being enabled by the user:

image3376

This shows that the port power turns on as soon as VAA is enabled, and stays on until `!ANT_BIAS` is raised ~150ms later.

#955 proposed a solution to this but didn't end up fixing the problem.
A change has been pushed to portapack-mayhem that is reported to fix it: https://github.com/eried/portapack-mayhem/pull/413.

### Version information
**Operating system**: Ubuntu 20.04

**hackrf_info output:**

```
hackrf_info version: git-c8695e0
libhackrf version: git-c8695e0 (0.6)
Found HackRF
Index: 0
Serial number: 0000000000000000909864c8345451cf
Firmware Version: git-26c0f70 (API:1.05)
Part ID Number: 0xa000cb3c 0x004e4f5c
Opera Cake found, address: 0
```

martinling:
This PR fixes #1006. The changes are:
- Move rad1o-specific delay out of `main` and into the rad1o version of `enable_rf_power`. This reduces the glitch from ~150ms to ~80ms, and speeds up startup on HackRF One.
- Change the default register settings for the RFFC5072 mixer to include driving the GPO pin for `!ANT_BIAS` high.
- On HackRF One, move the call to `mixer_setup` from `rf_path_init` to before `enable_rf_power` is called in `main`.

This gives the desired result: `!ANT_BIAS` is now driven high before `!VAA_ENABLE` is driven low. Note there is no load on the antenna port here, just a 1M scope probe: the measured voltage drifts upwards, presumably due to some leakage, but never gets pulled up hard to VAA and should remain near-zero with any load attached.

image


mossmann (approved):
Yay!

PR #1165, opened 2022-09-19
libhackrf: #define outdated board IDs

mossmann:
This is intended to solve the concern raised on #1155, making https://github.com/pothosware/SoapyHackRF/pull/36 unnecessary.


martinling (approved):
I can't think of any reason you need brackets around the definitions here, but otherwise this looks fine.

mossmann:
> I can't think of any reason you need brackets around the definitions here, but otherwise this looks fine.

Fair! I just bracket every #define because I've been burned before. :joy:

PR #1166, opened 2022-09-20
Fix hang in TX when a libusb error occurs.

### LINKED ISSUE #1163: [Bug Report]: `hackrf_transfer` hangs when device disappears in TX mode

If device is reset or unplugged during `hackrf_transfer -t` or `-c`, the command hangs and does not terminate.

- `hackrf_transfer -c 0`
- press RESET button during execution

_No response_

martinling:
Where our callback is called with a status other than `LIBUSB_TRANSFER_COMPLETED`, something abnormal has happened and we no longer want to either submit further transfers or await a flush. So clear both the `streaming` and `flush` flags, and proceed to transfer_finished.

Do the same if `libusb_submit_transfer` fails when called from our callback to submit the flush transfer.

This fixes #1163.


mossmann (approved):
Looks great!

PR #1168, opened 2022-09-20
Provide a callback for TX flush, rather than a wait function.

### LINKED ISSUE #1164: [Bug Report]: `hackrf_transfer` misbehaves with `-H -n 393216` and `-c` or `-t`

When `hackrf_transfer` is configured to wait for a trigger and then transmit a small number of samples (fewer than 393217), the "Waiting for sync" output does not appear. During this wait the command cannot be terminated with Ctrl-C.

- leave trigger input unconnected
- `hackrf_transfer -c 0 -H -n 393216`
- type Ctrl-C

_No response_

martinling:
This fixes #1164, but currently causes some erroneous output to be printed in `hackrf_transfer` e.g:
```
$ hackrf_transfer -c 0 -n 100000
call hackrf_set_sample_rate(10000000 Hz/10.000 MHz)
call hackrf_set_freq(900000000 Hz/900.000 MHz)
samples_to_xfer 100000/0Mio
Stop with Ctrl-C
 0.0 MiB / 0.011 sec =  0.0 MiB/second, average power -nan dBfs

Couldn't transfer any bytes for one second.
```

### DIFF
--- a/host/libhackrf/src/hackrf.c
+++ b/host/libhackrf/src/hackrf.c
@@ -1768,26 +1768,6 @@ static void* transfer_threadproc(void* arg)
 	return NULL;
 }
 
-static void transfer_finished(
-	struct hackrf_device* device,
-	struct libusb_transfer* finished_transfer)
-{
-	// If a transfer finished for any reason, we're shutting down.
-	device->streaming = false;
-
-	// If this is the last transfer, signal that all are now finished.
-	pthread_mutex_lock(&device->all_finished_lock);
-	if (device->active_transfers == 1) {
-		if (!device->flush) {
-			device->active_transfers = 0;
-			pthread_cond_broadcast(&device->all_finished_cv);
-		}
-	} else {
-		device->active_transfers--;
-	}
-	pthread_mutex_unlock(&device->all_finished_lock);
-}
-
 static void LIBUSB_CALL hackrf_libusb_flush_callback(struct libusb_transfer* usb_transfer)
 {
 	// TX buffer is now flushed, so proceed with signalling completion.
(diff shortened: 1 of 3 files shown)


martinling:
The message `Couldn't transfer any bytes for one second` appears only for `-n 393216` and below, which corresponds to three 256KiB transfers.

That seems to be an unrelated issue, related to the way stats were adjusted for the preload bytes in https://github.com/greatscottgadgets/hackrf/commit/8a9af7a1add97f82ed4b62d4d0b84e021e0a0219.

However, we can now Ctrl-C when `-H` is used with those `-n` values, which does fix #1164.

martinling:
Added an additional commit to fix the `hackrf_transfer` stats.

martinling:
Added several more commits to solve remaining problems with `hackrf_transfer` output, including the introduction of an additional callback for the completion of a TX transfer, which greatly simplifies the bookkeeping work needed on the client side.

host/libhackrf/src/hackrf.c
```diff
@@ -724,6 +727,8 @@ static int hackrf_open_setup(libusb_device_handle* usb_device, hackrf_device** d
 	lib_device->active_transfers = 0;
 	lib_device->flush = false;
 	lib_device->flush_transfer = NULL;
+	lib_device->flush_callback = NULL;
+	lib_device->flush_ctx = NULL;
```
mossmann, line 731: Should this also set do: `lib_device->tx_completion_callback = NULL;`?

martinling: Fixed.

martinling:
Added some further commits to ensure that the flush transfer is included in the cancellation and locking arrangements.

The flush callback now takes a success parameter to indicate if flushing was successful or cancelled/failed.

martinling:
Something about this branch breaks Ctrl-C on macOS. I'm investigating.

martinling:
Bisection shows the breaking commit is 9f7715e5.

martinling:
Now fixed; `resubmit` wasn't being initialised to `false` outside the `if` blocks.

mossmann (approved):
Wonderful! I find the logic in hackrf_transfer much easier to read and understand with this.

PR #1172, opened 2022-09-21
Fix initial state of RF path GPIOs.

### LINKED ISSUE #1169: [Bug Report]: TX amp is on by default at startup

Unless the TX amp is explicitly turned off with `-a 0`, the first run of `hackrf_transfer` after the HackRF is powered up will have the amp enabled.

0. Connect an oscilloscope or other measuring instrument to the RX output.
1. Plug in or reset HackRF.
2. Run e.g. `hackrf_transfer -c 127 -f 14e6 -x 30`
3. Check TX output level.
4. Stop `hackrf_transfer` with Ctrl-C.
5. Run the same command again.
6. Check TX output level which is now reduced by the amp gain.

We could work around this in `hackrf_transfer`, but other libhackrf clients would still be affected. Really the default firmware behaviour should be to keep the amp off unless enabled with the relevant USB vendor request.

I'm not sure where the error is in the initial state. The definition of `SWITCHCTRL_SAFE` in `rx_path.c` seems correct.

_No response_

martinling:
Previously `rf_path_pin_setup` would set up the GPIOs with:

`switchctrl_set(rf_path, SWITCHCTRL_AMP_BYPASS | SWITCHCTRL_MIX_BYPASS)`

However, this misses out `SWITCHCTRL_NO_TX_AMP_PWR` and `SWITCHCTRL_NO_RX_AMP_PWR` which are needed to keep the amps off. The correct setting to use is defined as `SWITCHCTRL_SAFE`. The call was also made after setting the GPIOs to output mode, so there could be a brief glitch.

The `SWITCHCTRL_SAFE` value was being assigned to the global `switchctrl` setting but this was only used once, in `rf_path_init`, which does apply that state shortly after `rf_path_pin_setup`. However, all other functions use `rf_path->switchctrl`, not the global `switchctrl`, which was never being initialised at all.

This PR corrects these issues, which fixes #1169.

### DIFF
--- a/firmware/common/rf_path.c
+++ b/firmware/common/rf_path.c
@@ -341,7 +341,7 @@ void rf_path_init(rf_path_t* const rf_path)
 #ifndef HACKRF_ONE
 	mixer_setup(&mixer);
 #endif
-	(void) rf_path; /* silence unused param warning */
+	rf_path->switchctrl = SWITCHCTRL_SAFE;
 }
 
 void rf_path_set_direction(rf_path_t* const rf_path, const rf_path_direction_t direction)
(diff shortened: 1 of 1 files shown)


mossmann (approved):
Excellent!

PR #1180, opened 2022-09-26
Improve transfer locking

martinling:
It was possible in theory for a transfer to complete before `prepare_transfers` had finished initially submitting transfers, leading to various potential races or out-of-order data.

This PR prevents this by extending the scope of `transfer_lock` to more of the libusb transfer callback, and taking this lock during `prepare_transfers` until all initial submissions are complete.

After this change, it became possible to merge `all_finished_lock` with `transfer_lock`, which simplifies things significantly.


mossmann (approved):
Very nice!

PR #1205, opened 2022-10-24
Docs: Move Opera Cake documentation to its own section

straithe:


docs/source/index.rst
```diff
@@ -4,11 +4,9 @@ Welcome to HackRF's documentation!
 
 .. toctree::
    :maxdepth: 2
-   :caption: User Documentation
+   :caption: HackRF User Documentation
```
mossmann, line 7: I'm confused by the name change to "HackRF User Documentation", especially with Jawbreaker being moved out of this section. It kind of looks like you meant it to be "HackRF One" not "HackRF", but a lot of the content in this section is relevant to all three HackRF platforms, not just HackRF One. The word "HackRF" does not appear in the name of any other section even though they are all directly related to HackRF (except arguably the Opera Cake section).

docs/source/index.rst
```diff
@@ -46,3 +44,20 @@ Welcome to HackRF's documentation!
...
+   opera_cake_hardware
+   opera_cake_board_addressing
+   opera_cake_port_configuration
+   opera_cake_modes_of_operation
+
+.. toctree::
+   :maxdepth: 2
+   :caption: Jawbreaker Documentation
+
+   jawbreaker
```
mossmann, line 63: I don't like moving the Jawbreaker page away from the HackRF One page unless you plan to also move HackRF One to its own section. At a later time I think we should add a page for rad1o which is the third HackRF platform supported by the software in this repo. I would like either each platform to have its own section or all three platforms to be in the same section.

I notice that the "Hardware" section is almost entirely about HackRF One and could be converted to a HackRF One section. If you are okay with doing that in the future, then I am okay with moving Jawbreaker here.

mossmann (changes requested):
I love the breakout of the Opera Cake content, but I would like to see a change or two.

docs/source/opera_cake_hardware.rst
```diff
@@ -0,0 +1,18 @@
+========
+Hardware
```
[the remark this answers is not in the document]
straithe: Thanks!

straithe:
Changes made! Thank you for the review.

mossmann (approved):
Wonderful! Thank you!

PR #1212, opened 2022-10-26
m0: reset M0 before starting it

schneider42:
The rad1o was not starting the M0 when powered up by inserting a USB cable. Interestingly the M0 does start when toggling the power switch.

Resetting the M0 before starting it in `main()` solves this issue.

### DIFF
--- a/firmware/hackrf_usb/hackrf_usb.c
+++ b/firmware/hackrf_usb/hackrf_usb.c
@@ -241,6 +241,7 @@ int main(void)
 	cpu_clock_init();
 
 	/* Wake the M0 */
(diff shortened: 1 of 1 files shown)


martinling (approved):
It hadn't ever occured to me that the M0 could be in any state other than halted at startup, so I feel like there may still be something else going on behind this. But I don't see any problem with adding the halt call here, and it fixes the problem, so I think we should merge this.

schneider42:
I don't have an explanation for the behaviour. I can only assume that the rise time of the power supply or something similar has an influence here.

PR #1217, opened 2022-10-31
Remove default titles from issue forms

miek:
This removes the pre-populated titles from the issue form templates. I think in practice they end up being confusing, and users often either overwrite them or leave the title empty. Since the template adds labels to the issues anyway, I think we can remove the title tags.


straithe (approved):
Looks great. Thank you!

Issue #1230, opened 2022-11-16
[Bug Report]: Inconsistent bin results from hackrf_sweep one-shot mode


grvvy:
### What type of issue is this?

permanent - occurring repeatedly

### What issue are you facing?

The target signal power of ~ -20dB should always appear in the second bin of this test output but it somewhat regularly appears shifted into a different bin or totally out of scope at either/both 9.5MHz and 2665.5MHz. Meanwhile 915.5MHz and 5995.5MHz are consistently rock solid.

Changing the sweep call to sweep twice instead of using one-shot mode results in the 1st sweep failing as does here, but the 2nd sweep never fails along with the 1st.

### What are the steps to reproduce this?

https://gist.github.com/grvvy/a1a9c970d05a6c8654bd1fd2aa579e34

Running this script with two hackrf's plugged in with antenna and clk ports connected.

### Can you provide any logs? (output, errors, etc.)

The test fails here after the 4th run (signal not appearing at 9.5MHz and signal (-20.73) shifted to the 9th bin at 2665.5MHz).
```
Running test...
2022-11-16, 15:49:23.103952, 9000000, 14000000, 333333.33, 60, -35.83, -23.87, -24.87, -39.90, -54.86, -71.27, -64.26, -67.41, -66.41, -72.92, -68.81, -65.44, -82.04, -66.58, -64.18
2022-11-16, 15:49:24.219668, 915000000, 920000000, 333333.33, 60, -28.07, -22.44, -28.43, -66.44, -74.80, -85.79, -77.05, -70.29, -64.99, -65.41, -65.00, -72.53, -76.89, -80.10, -80.51
2022-11-16, 15:49:25.336927, 2665000000, 2670000000, 333333.33, 60, -23.36, -17.69, -23.61, -57.42, -60.59, -63.11, -62.20, -64.62, -65.04, -69.21, -70.05, -69.54, -69.92, -86.45, -77.35
2022-11-16, 15:49:26.437785, 5999000000, 6004000000, 333333.33, 60, -26.17, -20.57, -27.03, -49.25, -41.32, -45.67, -49.85, -52.20, -51.93, -51.13, -59.96, -44.10, -42.67, -48.11, -46.92
Running test...
[...]

PR #1231, opened 2022-11-17
Add testing for hackrf_transfer to CI

grvvy:


martinling (approved):
This looks great! I don't know what's going on with #1230 yet, but the workaround makes sense.

PR #1233, opened 2022-11-23
CI: add hackrf_debug test

grvvy:


martinling (approved):
Looks good to me!

mossmann:
FYI, this will require an update for r9 which doesn't have MAX2837.

PR #1235, opened 2022-11-30
Make definition of "prev" consistent

stevefalco:
The definition of `prev` is `r1` at line 360 in macro `handle_shortfall`, but is later changed to `r0` in `checked_rollback` at line 569.  This causes a warning message:

`sgpio_m0.s:569: Warning: ignoring redefinition of register alias 'prev'`

Doing a disassembly shows that the assembler indeed ignores the redefinition and uses `r1` in `checked_rollback`.

This pull request simply changes the definition in `checked_rollback` to match what macro `handle_shortfall` is using, and thus eliminates the warning without actually changing the binary code that is generated.

### DIFF
--- a/firmware/hackrf_usb/sgpio_m0.s
+++ b/firmware/hackrf_usb/sgpio_m0.s
@@ -566,7 +566,7 @@ checked_rollback:
 	beq idle                                        //      goto idle                       // 3
 
 	// Otherwise, roll back the state to ignore the current shortfall, then jump to idle.
-	prev .req r0
+	prev .req r1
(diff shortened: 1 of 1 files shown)


martinling (approved):
Thanks for spotting and fixing this! I thought I had already fixed these warnings during the development of #982, but I guess this one slipped through one of the rebases.

(It is rather annoying that registers can only be given names like this on a file-wide scope in the first place, but I thought it was worthwhile naming everything for readability nonetheless, given that it was straightforward enough to find a set of assignments that worked without needing to redefine them, give or take having slipped up on this one)

PR #1237, opened 2022-12-09
restrict IF to 2170-2740 MHz

mossmann:
This fixes a bug on certain HackRF Ones and Jawbreakers with a MAX2837 or MAX2839 that fails to tune to as wide a range as most other units. This screenshot shows a TX sweep of two HackRF One r4 (A and F) and four HackRF One r9 with at least one of each revision exhibiting the problem:

Copy20221208_003


martinling (approved):
This all looks correct to me, and I can't find any places that you've not updated the range.

PR #1238, opened 2022-12-11
Add hackrf_tcp tool

VVakko:
Added the `hackrf_tcp` tool.

The compilation and working of this tool has been tested on Linux (Ubuntu 20.04/22.04 on x64, Ubuntu 20.04 on RPi4), Windows 10 (Visual Studio 2022, x86/x64) and macOS (Monterey).

In normal mode, it is fully compatible with the `rtl_tcp` tool, the IQ stream from HackRF is converted to uint8 for correct working in programs like **GQRX**, **SDR++**, etc. But this conversion additionally loads the CPU. This is also why the extended mode was added by analogy with `rsp_tcp` (option `-E`). In this mode, the IQ stream is not converted and is transmitted in int8 format. Also, thanks to the transmitted header, clients can recognize **HackRF** and additionally enable samplerates, which **RTL-SDR** does not support, but **HackRF** supports.

I usually use a MacBook, and to run **SDR#** I use **Parallels Desktop**. But **Parallels Desktop** does not work well with SDR receivers connected via USB. Due to jitter, on large samplerates, there is a voice trembling and it becomes impossible to decode digital transmissions. If you do not throw USB devices into a virtual machine, but run `hackrf_tcp` on macOS natively, then you can work with any samplerate in a virtual machine via tcp.

HackRF


straithe:
Thank you for your pull request. We do think it is good work but we will not be accepting. We are not looking to maintain this tool in this repository, but we do encourage you to host it in your own repository.

PR #1239, opened 2022-12-13
Add Windows build of libhackrf and host tools to Github Actions

martinling:
This PR adds a Windows MSVC build of libhackrf and the host tools to our existing Github Actions workflow. The aim is to replace the current Appveyor builds, which use outdated dependencies and require a separate CI platform.

The build uses the current GitHub `windows-latest` runner. The dependencies (libusb, pthreads and fftw) are first installed via [vcpkg](https://github.com/microsoft/vcpkg/). The latest versions currently available through the installed vcpkg will be used.

Builds are then configured, passing all the necessary header and library paths to CMake's configure step, which generates MSBuild project files. Ideally, explicitly specifying the paths should be eliminated in future by having our CMake scripts integrate better with vcpkg, but the current approach works for now without changes to our build system.

As with the Linux and macOS builds, first libhackrf and the host tools are built together, then libhackrf is built on its own and installed. The host tools are then configured to be built separately against the installed copy of libhackrf.

However, the final build of the host tools against the installed libhackrf is skipped, because MSVC fails to find `hackrf.h`. I don't currently know why. The header is installed where it should be, and the configure step succeeds with the `LIBHACKRF_INCLUDE_DIR` set to that location. I was not able to reproduce the failure locally with the same steps.

This doesn't seem like a significant problem however, as the combined build of the `host` directory still verifies that the code builds successfully under Windows and MSVC.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,9 +8,6 @@ on:
   schedule:
     - cron: 1 12 * * 1
 
-env:
-  BUILD_TYPE: Release
-
 jobs:
   host:
     strategy:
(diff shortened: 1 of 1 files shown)


mossmann (approved):
Wonderful! Thank you!

PR #1245, opened 2022-12-18
add --led option to hackrf_debug

mossmann:
and corresponding changes to libhackrf and firmware


martinling (approved):
Looks good to me. I might hang my rad1o on the christmas tree now!

PR #1251, opened 2023-01-04
Add firmware info structure

miek:
This adds a structure at a fixed location within the firmware image, to provide information about the firmware image.
In particular, it has a field to specify which hardware platforms the firmware image supports that `hackrf_spiflash` can use to warn about flashing incompatible firmware. This replaces the old method of searching the USB string descriptors and allows for a firmware image to support multiple platforms.

I've marked this as a draft for now as it depends on https://github.com/mossmann/libopencm3/pull/31 to define a new section for the struct. Once that's merged in, I'll add a submodule-update commit & rebase.

### DIFF
--- a/firmware/common/firmware_info.h
+++ b/firmware/common/firmware_info.h
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2022 Great Scott Gadgets <info@greatscottgadgets.com>
+ *
+ * This file is part of HackRF.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2, or (at your option)
+ * any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; see the file COPYING.  If not, write to
+ * the Free Software Foundation, Inc., 51 Franklin Street,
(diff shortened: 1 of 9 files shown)


host/hackrf-tools/src/hackrf_spiflash.c
```diff
@@ -102,11 +102,60 @@ int compatibility_check(uint8_t* data, int length, hackrf_device* device)
...
-	return 1;
+	return EXIT_FAILURE;
+}
+
+#define FW_INFO_LOCATION              0x400
+#define FW_MAGIC_OFFSET               0
+#define FW_STRUCT_VERSION_OFFSET      8
+#define FW_SUPPORTED_PLATFORMS_OFFSET 12
+
+#define FROM_LE32(x) ((x)[0] | ((x)[1] << 8) | ((x)[2] << 16) | ((x)[3] << 24))
```
mossmann, line 117: Thank you for your attention to endianness!

firmware/common/firmware_info.h
```diff
@@ -0,0 +1,36 @@
...
+#ifndef FIRMWARE_INFO_H
+#define FIRMWARE_INFO_H
+
+#include <stdint.h>
+
+struct firmware_info_t {
+	char magic[8];
+	uint32_t struct_version;
+	uint32_t supported_platform;
+	char version_string[32];
```
mossmann, line 31: Could you please add a field that indicates whether or not the firmware was compiled with DFU_MODE?

mossmann (changes requested):
This is lovely, but I would like to see one change.

firmware/common/firmware_info.h
miek: Done! I've reduced the size of the `struct_version` field and added `dfu_mode` next to it, to keep everything on nice boundaries.

mossmann (approved):
Excellent! Thank you!

PR #1253, opened 2023-01-04
Fix memory waste in allocated_transfers

Sasszem:
Did a search for allocations & all others were fine


martinling (approved):
This is correct. We were allocating space for `TRANSFER_COUNT` times `sizeof(struct libusb_transfer)`, but we only needed to allocate space for a pointer to each struct. The actual structs are allocated further down, by `libusb_alloc_transfer`.

mossmann:
@Sasszem Is there a reason this is marked as a draft? Is it okay if we go ahead and merge it?

mossmann:
Thank you!

PR #1283, opened 2023-02-14
Add files via upload

goatgitt:
Hello,
I've made the all pcb in two layers including ground plane from the 2nd and 3rd layer.
I'm new in the project so there is maybe some mistakes out there.

### DIFF
--- a/hardware/hackrf-one/hackrf-one.kicad_pro
+++ b/hardware/hackrf-one/hackrf-one.kicad_pro
@@ -377,7 +377,7 @@
     "pinned_symbol_libs": []
   },
   "meta": {
-    "filename": "hackrf-one-modified.kicad_pro",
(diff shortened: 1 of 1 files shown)


straithe:
Can you please add a picture of your printed PCB?

goatgitt:
I've done this yesterday so I don't have it yet but I did some checks( DRC, numerical simulations) But in a physical way, I'm not shure if it is working at 100% because I don't have the material for further testing (like Oscilloscope and radio equipment)

straithe:
We've talked internally about this PR and won't be accepting it. We aren't looking to maintain a 2-layer version of HackRF in this repository. If you wish to keep going with this project on your own we suggest that you look into the characteristic impedance needed by traces in RF devices. You may also find some useful information in https://www.youtube.com/watch?v=TnRn3Kn_aXg useful

Issue #1311, opened 2023-04-23
stand alone Bias-T control utility


nmoon88:
### What feature would you like to see and why?

I really would like an external utility to enable and disable bias-t for the HackRF One and similar devices. Being forced to rely on certain programs to incorporate it severely limits what can be done and such a utility already exists for RTL-SDR sticks. This would be useful on Windows, Linux and MacOS or any other platform that supports the device.

ai6aj:
I've submitted a pull request to add this to the main repo but if you need the utility ASAP you can find it in my fork: https://github.com/ai6aj/hackrf/tree/20230418_suitejo_biast

One caveat, the HackRF firmware turns off the bias T power whenever it leaves RX mode.  So once you hit STOP or exit SDR# (or another program) the external power will shut off and you'll need to re-enable it.  IMO it's still much less of a PITA than using external power for your LNA or other device.

martinling:
We have two issues open for the same thing here. Let's continue the discussion on #1309. I'll close this one.

Issue #1309, opened 2023-04-19
Command line tool to switch antenna power on/off, similar to rtl_biast


ai6aj:
### CLI tool for Antenna Power control

Some software (notably SDR#) does not provide a way to turn on antenna power to power an offboard LNA.  On RTL-SDR the easy workaround is to use rtl_biast to enable the power circuit, then start your software:

$ rtl_biast -b 1

I cannot find a method of doing this for HackRF.  I have written hackrf_biast.c and placed it in my fork but won't submit a pull request if there's an official (or just better) way of doing this.

nmoon88:
I requested exactly the same thing last night. It seems like this is something that would benefit everyone but the developers don't want to do it. I was told to hire a C developer to code such a utility in the Discord chat.

martinling:
This can't be done purely with a host-side utility, it would also need firmware changes. Currently the firmware turns off the bias tee every time it leaves TX or RX mode.

ai6aj:
> This can't be done purely with a host-side utility, it would also need firmware changes. Currently the firmware turns off the bias tee every time it leaves TX or RX mode.

As long as the power stays on to _start_ receiving, that's good enough.  I have written the utility and submitted a pull request, antenna power stays on in SDR# until you hit STOP or exit the program... at which point you need to re-enable it again.  But that's a lot better than not being able to turn it on at all.

martinling:
I had a quick chat with @mossmann about this, and we think a useful direction for development would be to add a bias tee control mode setting to the firmware, which would have multiple options for how the bias tee would be controlled:

- Host controlled (default, current behaviour).
- Turn on automatically during RX only.
- Turn on automatically during TX only.
- Turn on automatically during both TX and RX.
- Turn on and leave on.
- Turn off and leave off.

This would provide better control over how the bias tee should be used in software that doesn't control it directly. It would also be useful even for software that's aware of the bias tee, since it would be possible to speed up and simplify mode changes by telling the firmware to control the bias tee automatically, rather than having to request changes each time.

There would need to be a new USB vendor request type and a new libhackrf function to set the bias tee control mode. Then a command line tool like @ai6aj's `hackrf_biast` proposal could be written to control it.

PR #1312, opened 2023-04-24
hackrf_biast - (temporarily) enable antenna power from the command line

ai6aj:
Adds a simple utility, hackrf_biast, to enable the antenna power from the command line for use with software such as SDR# which only has the bare minimum support for HackRF.  Addresses issues #1309 and #1311

Caveat: the antenna power will turn off once RX is stopped


ai6aj:
#1309

martinling:
Having it only work the first time you start RX isn't really a solution. I think that if we merge this we will just get a lot of bug reports saying it doesn't work properly.

If we're going to have utilities that override HackRF settings, to work around issues in closed source software, then that really needs firmware support so that the forced settings can be made persistent.

It would be a lot simpler for everyone if SDR# would just add a control for the bias tee.

ai6aj:
> If we're going to have utilities that override HackRF settings, to work around issues in closed source software, then that really needs firmware support so that the forced settings can be made persistent.

Agreed.  It's been a few years since I hacked on the firmware but I don't think it's particularly difficult.  I'll

martinling:
If you want to work on the firmware support that'd be great. I've added some thoughts on how this could work [here](https://github.com/greatscottgadgets/hackrf/issues/1309#issuecomment-1521504784).

PR #1314, opened 2023-04-28
Firmware updates to allow fine-grained control of biast from command line

ai6aj:
Firmware patch and command line utility to address #1309.  Bias tee power can be set to "on", "off", or "device default" for each of the RF path modes (TX, RX, OFF).  Example usage would be to have bias tee power turn ON by default whenever the device enters RX mode.

The command line tool _hackrf_biast_ supports turning on bias tee power immediately as well as configuring settings for one or more RF paths.

@mossmann suggested that it would be useful (and faster) for host software to be able to specify bias t settings once instead of invoking the API to reconfigure the device every time it is switched from RX to TX etc.  It seems like it would also be useful to do this for LNA settings etc. so I placed the bias t code in a new file named 'user_config.c' rather than just hacking it into rf_path.c.

User-supplied bias tee config is applied after all other RF path switching has completed.  This can cause a very brief glitch in bias tee power when switching modes if the device's default action is to turn the power off.  Given that prior behavior would've produced a much longer power shutoff it shouldn't be a problem.


martinling (changes requested):
I like this, it's a neat solution.

I think it can be made simpler and clearer by having mode `00` mean "do nothing" rather than "device default".

There's nothing particularly special about the default behaviour; it can be expressed within your new configuration mechanism (TX: do nothing, RX: do nothing, OFF: disable bias-T). So there's no need for a special way to request that combination. We can just make that the default combination at reset, and have `hackrf_biast -R` restore it.

There's a bit more work to do in `rf_path_set_direction`: 

- You've not removed the code which disables the bias-T on transition to `OFF`, so at the moment a `NOP` configuration for `OFF` mode won't actually behave as such. That call can be removed once the default configuration is changed such that `user_config_on_rf_path_direction_change` will turn the bias-T off in that transition.

- Some thought is needed about what order things should happen in during RX path changes. Currently when entering `OFF` mode the bias-T is disabled before other changes. When software enables the bias-T, it generally does so before starting TX/RX.  So I think it's more appropriate to move the `user_config_on_rf_path_direction_change` above the switch statement.

Aside from that, I think this is good to go, but I'd like @mossmann's thoughts.

firmware/common/user_config.c
```diff
@@ -0,0 +1,100 @@
...
+ * the Free Software Foundation, Inc., 51 Franklin Street,
+ * Boston, MA 02110-1301, USA.
+ */
+
+#include "user_config.h"
+
+
+static user_config_user_opt_t user_direction_rx_bias_t_opts = RF_DIRECTION_USER_OPT_NOP;
+static user_config_user_opt_t user_direction_tx_bias_t_opts = RF_DIRECTION_USER_OPT_NOP;
+static user_config_user_opt_t user_direction_off_bias_t_opts = RF_DIRECTION_USER_OPT_NOP;
```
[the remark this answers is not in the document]
ai6aj: Ok, I've made all the changes you requested, should be good to go.

host/hackrf-tools/src/hackrf_biast.c
```diff
@@ -0,0 +1,180 @@
+/*
+ * Copyright 2012-2022 Great Scott Gadgets <info@greatscottgadgets.com>
+ * Copyright 2012 Jared Boone <jared@sharebrained.com>
+ * Copyright 2013 Benjamin Vernoux <titanmkd@gmail.com>
```
martinling, line 4: None of these people wrote this code! You should put your own details here (and similarly on other files you've added).

martinling (changes requested):
Thanks for the fixes. I've flagged up a few remaining things inline.

I also see another issue: having changed the meaning of `00` to `do nothing` and the bitmask applied by `-R` to `0b100100110`, it's no longer quite correct to simply use `|=` to update `user_biast_modeopts` in `hackrf_biast`.

Consider the case of `hackrf_biast -R -o 0`: reset to defaults, but leave the bias tee alone when entering OFF mode.

Bit 1 is now set when handling the `-R` option, but not cleared when handling the `-o` option, so the `-o 0` part ends up having no effect.

That's easy to fix by clearing old mode bits in `update_user_mode`. But I do wonder whether it would be better to avoid all this bit twiddling in `hackrf_biast`, and by extension, any other code that wants to use this feature.

It makes sense to pack all these settings into a `uint16` at the USB request level, but it might be a lot clearer if the libhackrf API used a struct instead, perhaps something like:

```c
struct hackrf_biast_setting {
        bool modify;
        bool change_on_mode_entry;
        bool enable;
}

struct hackrf_biast_request {
        hackrf_biast_setting tx;
        hackrf_biast_setting rx;
        hackrf_biast_setting off;
}       
```

That corresponds exactly to the bitmask used on the wire, but would be a lot clearer and less error-prone to use.

Another way that usability could be improved is to use words rather than numbers for the `-t`, -`r` and `-o` options to `hackrf_biast`. Currently `1` means "turn off", which is quite counterintuitive and I think people are very likely to get that wrong. If words are used instead the meaning is much clearer and harder to get wrong, e.g. `hackrf_biast -r on -t off -o leave`.

host/hackrf-tools/src/hackrf_biast.c
ai6aj: Forgetting to put your name on things is how you know you've been working on proprietary code way too long ;)

ai6aj:
Ok, those changes have been made.  Hopefully third time's the charm.

ai6aj:
The format issue in usb_api_register.h was my fault, but clang-format is perfectly happy with these lines on my system and it's not my code...

```
 host/libhackrf/src/hackrf.c:42:14: error: code should be clang-formatted [-Wclang-format-violations]
        #define true  1
                    ^
host/libhackrf/src/hackrf.c:140:26: error: code should be clang-formatted [-Wclang-format-violations]
        volatile bool streaming;         /* volatile shared between threads (read only) */
```

martinling:
Hm, something weird going on there. What version of clang-format are you running?

I've got:
```
$ clang-format --version
Debian clang-format version 14.0.6
```
...and I currently get the same result as the CI run, i.e. it just wants to move the comment on the `streaming` line.

ai6aj:
I'm using clang-format v.16, I can't find a prebuilt v14 for Raspbian.

Strange that removing a couple spaces before a comment caused this from Jenkins:

...
+ python3 ci-scripts/test-transfer.py rx
2023-05-03, 16:16:25.777018, 9000000, 14000000, 333333.33, 60, -69.47, -69.69, -81.41, -68.49, -72.57, -65.17, -66.25, -72.92, -68.13, -85.88, -65.24, -64.02, -67.01, -70.13, -67.93
2023-05-03, 16:16:26.874496, 915000000, 920000000, 333333.33, 60, -27.34, -21.61, -27.44, -59.90, -64.46, -70.54, -74.18, -72.29, -73.39, -71.13, -75.36, -75.55, -70.76, -67.98, -80.10
2023-05-03, 16:16:27.980753, 2665000000, 2670000000, 333333.33, 60, -23.65, -17.58, -23.17, -55.24, -65.93, -75.19, -79.59, -74.30, -77.26, -70.57, -70.69, -74.37, -71.10, -74.72, -77.04
2023-05-03, 16:16:29.085658, 5999000000, 6004000000, 333333.33, 60, -27.31, -21.51, -27.21, -45.73, -40.41, -43.62, -46.69, -45.61, -57.74, -57.98, -53.38, -45.97, -44.20, -48.65, -47.42
Signal not strong enough at 9.5 MHz
Non-target bin power exceeded max power threshold at 9.5 MHz
...

Is something weird going on with the Jenkins setup?

martinling:
I think that test is behaving intermittently due to bug #1230. It's due to be replaced with a better test when #1285 is merged.

Now the Appveyor build seems to be broken, which isn't your fauilt either. I'd like to just get rid of that action, it's out of date and using ancient depenencies. Since #1239 we've had Windows builds via GitHub Actions instead.

ai6aj:
Looks like it was just a network error:

```
curl -fsS -o "C:\fftw-3.3.5.zip" "ftp://ftp.fftw.org/pub/fftw/fftw-3.3.5-dll64.zip"
curl: (6) Could not resolve host: ftp.fftw.org
Command exited with code 6
```

I just ftp'd into that host and it's up now, I'm going to try closing and reopening the PR to retrigger the build.

ai6aj:
Ok, that was the problem.  Is there anything left for me to do on this PR?

host/hackrf-tools/src/hackrf_biast.c
```diff
@@ -0,0 +1,222 @@
...
+		"  [-t mode]  Set default bias tee power when device enters TX mode\n");
+	fprintf(stderr,
+		"  [-o mode]  Set default bias tee power when device enters OFF mode\n");
+
+	fprintf(stderr,
+		"  [-d serial_number]  Specify serial number of HackRF device to configure\n\n\n");
+	fprintf(stderr, "The -r/-t/-o options support the following mode settings:\n\n");
+	fprintf(stderr, "  leave		do nothing when entering mode\n");
+	fprintf(stderr, "  on		disable bias tee when entering mode\n");
+	fprintf(stderr, "  off		enable bias tee when entering mode\n\n");
```
martinling, line 64: The names and meanings are out of sync here.

host/hackrf-tools/src/hackrf_biast.c
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	setting->do_update = true;
+	if (mode < 2) {
+		setting->change_on_mode_entry = true;
+		setting->enabled = mode;
+	} else {
+		setting->change_on_mode_entry = false;
+		setting->enabled = false;
+	}
```
martinling, line 94: Now that the program is no longer taking numeric modes on the command line, I think this function would be much clearer if you just removed the `mode` variable entirely and just put all the effects in one if/else/else block, with a catchall `else` at the end for invalid values.
```suggestion
	if (strcmp("off", strarg) == 0) {
	        setting->do_update = true;
		setting->change_on_mode_entry = true;
		setting->enabled = false;
	} else if (strcmp("on", strarg) == 0) {
	        setting->do_update = true;
		setting->change_on_mode_entry = true;
		setting->enabled = true;
	} else if (strcmp("leave", strarg) == 0) {
	        setting->do_update = true;
		setting->change_on_mode_entry = false;
		setting->enabled = false;
	} else {
		fprintf(stderr, "Invalid mode '%s' for %s\n", strarg, direction_str);
		exit(INVALID_BIAST_MODE);
	}
```

martinling (changes requested):
There's one error that's crept into the usage text, and I have a suggestion to simplify the `update_user_mode` code, but other than that I'm happy with this.

I've asked @mossmann to have a look and confirm if he's happy with the API and CLI.

martinling (approved):
Looks good to me.

mossmann:
Thank you very much for the contribution! I have squashed and merged.

PR #1316, opened 2023-05-03
docs: move connectors information to its own section

straithe:


epenelope (approved):
These changes look good, @straithe!

Issue #1323, opened 2023-05-15
Broken timing in hackrf_transfer on OpenBSD, due to SIGALRM being delivered to transfer thread


bsdmp:
### What type of issue is this?

permanent - occurring repeatedly

### What issue are you facing?

while doing simple `hackrf_transfer -r /dev/null` I face following error

`Couldn't transfer any bytes for one second.`

full log:

```
call hackrf_set_sample_rate(10000000 Hz/10.000 MHz)
call hackrf_set_hw_sync_mode(0)
call hackrf_set_freq(900000000 Hz/900.000 MHz)
Stop with Ctrl-C
 7.9 MiB / 1.006 sec =  7.8 MiB/second, average power -28.5 dBfs
 7.6 MiB / 1.000 sec =  7.6 MiB/second, average power -28.4 dBfs
 7.6 MiB / 2.000 sec =  3.8 MiB/second, average power -28.5 dBfs
 0.0 MiB / 1.000 sec =  0.0 MiB/second, average power -nan dBfs

Couldn't transfer any bytes for one second.

Exiting... hackrf_is_streaming() result: streaming stopped (-1003)
Total time: 5.00616 s
hackrf_stop_rx() done
hackrf_close() done
hackrf_exit() done
fclose() done
exit
```

`v2021.03.1` works fine though, I have done a bisect and it came up to the following commit:

```
a09e9a20ed42567b1fe979ab4db12f9f2590b2a4 is the first bad commit
commit a09e9a20ed42567b1fe979ab4db12f9f2590b2a4
Author: Martin Ling <martin-git@earth.li>
Date:   Fri Aug 5 15:43:25 2022 +0100

    Overhaul timing in hackrf_transfer.

    Rather than using sleep() for 1s at a time, set up an interval timer
    that will fire once per second, and wait in the main loop for either
    this or some other event.

    On POSIX, the timing is set up with setitimer(), which generates a
    SIGALRM signal each time the timer fires. The main loop runs pause() to
    wait for any signal.

    On Windows, the timing is set up using CreateWaitableTimer, which
    provides an event handle that is set each time the timer fires. The main
    loop runs WaitForMultipleObjects() to wait on this and an interrupt
    event.

    The TX and RX callbacks can now stop the main loop immediately when they
    stop streaming. This fixes #1019.

 host/hackrf-tools/src/hackrf_transfer.c | 71 ++++++++++++++++++++++++++++++---
```

### What are the steps to reproduce this?

[...]

PR #1330, opened 2023-05-22
mask all signals in libhackrf transfer_threadproc

bsdmp:
hackrf_transfer uses pause() and SIGALRM to print statistics and POSIX doesn't specify which thread must recieve the signal, block all signals in transfer_threadproc(), so we don't interrupt their reception by hackrf_transfer or any other app which uses the library (#1323)


martinling (changes requested):
The code that modifies the signal mask will need to be wrapped in an `#ifndef _WIN32` so it doesn't break the Windows build.

You can run `./tools/reformat-source.sh` to make clang-format happy.

martinling (approved):

mossmann:
Thank you for the contribution! I have squashed and merged.

PR #1345, opened 2023-07-05
libhackrf: use stdint integer types

Sasszem:
I went through `libhackrf/hackrf.h` and `libhackrf/hackrf.c` and replaced 'normal' integer types with stdint.h int types to make them specific in size.
This should not break any compatibility, as it's only a name change.
I left string types unchanged as `char*` since they should be handled as strings, not as integer arrays.

### DIFF
--- a/host/libhackrf/src/hackrf.c
+++ b/host/libhackrf/src/hackrf.c
@@ -141,9 +141,9 @@ struct hackrf_device {
 	void* tx_ctx;
 	volatile bool do_exit;
 	uint8_t buffer[TRANSFER_COUNT * TRANSFER_BUFFER_SIZE];
-	bool transfers_setup;           /* true if the USB transfers have been setup */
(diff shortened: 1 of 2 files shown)


martinling:
It might arguably have been better to use explicitly-sized types in the API from the start, but now that the library is in use we can't change the types retroactively.

In C the choice of which types have which sizes is up to the platform and the compiler, so although e.g. an `int` and an `int32_t` might be the same thing on some platforms, that will not necessarily be the case everywhere. Substituting one for the other would be a backwards-incompatible change on platforms where the types don't match.

In practice it's possible that all the types you've substituted are equivalent on all the platforms that libhackrf can be built for in practice. I'm not sure whether that's actually true, but it could be. But if it is, then although this change would turn out to be safe, it's also unnecessary: because if the platforms all agree on the size of the generic types, specifying the sizes doesn't actually add any information for anyone.

So either way, this isn't a good idea, and we're not going to merge this change.

Issue #576, opened 2019-01-16
Document libhackrf functions


dominicgs:
Years ago we started some documentation for libhackrf functions on the wiki: https://github.com/mossmann/hackrf/wiki/libHackRF-API

It would be useful for people writing software that uses libhackrf to have documentation on the functions.  These should be written on the wiki

┆Issue is synchronized with this [Basecamp todo](https://3.basecamp.com/5295633/buckets/26082705/todos/4597414381) by [Unito](https://www.unito.io)

PR #1346, opened 2023-07-05
libhackrf: document required library versions

Sasszem:
As discussed in #576, required USB API versions (checked via `USB_API_REQUIRED` in `libhackrf/hackrf.c`) should be documented for all functions.
I added a list of the functions requiring a minimum USB API version in the doc, plus a note in every function's doc about the USB API it requires.


mossmann:
Thank you very much!

Issue #863, opened 2021-03-23
hackrf_sweep: add fftw wisdom support


gozu42:
(mainly a tracking issue to keep my thoughts/code and potential feedback on this in one place, not sure i can tag it enhancement myself) 

currently hackrf_sweep uses fttw plans generated at startup, in MEASURE mode.
this leads to serious delays for some binwidth on some platforms (raspi).
for an example binwidth of 10kHz (fftsize 2004) the delay is ~33sec on a raspi, on a pc the same only takes 200ms.

fftw has a "wisdom" mechanism that allows saving fft plans to disk.
it handles "does this wisdom work for the requested fft config" on its own, so the same wisdom file could be used even when changing binwidth.

while at it, i would like to expose the different planner flags (estimate, measure, patient, exhaustive) as a commandline option.
and also add an option that prints the current plan. (fftw provides a mechanism for that)

the fftw wisdom api is a bit awkward, so the only way i see is to first try to plan with the WISDOM_ONLY flag, which returns NULL if no matching plan is found in wisdom, and in that case plan again without WISDOM_ONLY and write the now-new wisdom out to the given file.

a commandline way to study the planner performance outside of hackrf_sweep is: time fftwf-wisdom -m cof2004

gozu42:
trying to summarize the discussion we just had about this on discord.

* a commandline option to set a wisdom file name like "-W wfn"
* the wfn default to something in platform-specific tempdir 
* defaulting to tempdir means users are less likely to end up with an outdated/bogus wisdom file
* it also means after each reboot (or other clearing of tempdir) the first sweep will be "slow" again
* handle errors during save+load gracefully (there is just too much that can go wrong in all kinds of weirds way, just print an oops and move on, but dont fail hard)

PR #1368, opened 2023-10-27
Run fftw_execute() once on an empty buffer (Fix #1366), add option to save wisdom and specify plan (#863)

### LINKED ISSUE #1260: hackrf_sweep sometimes doesn't work on Raspberry Pi Zero 2 W

After about five to ten times of running the hackrf_sweep function on the Raspberry Pi Zero 2 W, it stops working and outputs:
0 total sweeps completed, 0.00 sweeps/second

My question is, if it's supposed to work on the Raspberry Pi Zero 2 W.
(I have tried the same thing on the Raspberry Pi 4 B, and it doesn't work.)

run   $ hackrf_sweep -f 10:6010 -B -w 1000000 -l 16 -g 24 -1    repeatedly

$ uname -a
Linux raspberry pi 5.10.103-v7+ #1529 SMP Tue Mar 8 12:21:37 GMT 2022 armv7l GNU/Linux

$ hackrf_info
hackrf_info version: git-b6acd66c
libhackrf version: git-b6acd66c (0.7)
Found HackRF
Index: 0
Serial number: 00000000000000000f77c60dc279182c3
Firmware Version: git-6f26b74e (API:1.07)
Part ID Number: 0xa000cb3c 0x00544754
Hardware Revision: r7
Hardware appears to have been manufactured by Grat Scott Gadgets.
Hardware supported by installed firmware:
	HackRF One

### LINKED ISSUE #1366: `hackrf_sweep` should ensure FFTW plan is ready before starting RX

The `hackrf_sweep` tool should run the FFT once on an empty buffer at startup, before actually starting the sweep, and discard the result. That would get the planning out of the way before actual data starts flowing, rather than having the first FFT call delay the USB callback whilst FFTW plans and sets up the FFT implementation. On slow hosts with large FFTs, this delay can be enough to cause a timeout.

I expect this will fix #1260.

See also #863, which would allow caching that work between runs. However, in the initial case where there is no cached plan, we would still need this change to avoid a timeout.

See #1260.

_No response_

ai6aj:
Runs fftw_execute() once on an empty buffer to make sure the plan
is generated before streaming begins.  Fixes #1366 and should fix #1260.

Implements #863, with some changes:

1. Adds a -W option to specify a wisdom file.  If the necessary wisdom does not exist (either because the file does not exist, or an equivalent or better plan does not exist in the file) we attempt to save it at program exit.
   
2. Adds a -P option to specify the planner flags (estimate, measure, patient, exhaustive).
   
3. There is no reason to use the WISDOM_ONLY flag; FFTW will automatically use the best plan it can find in the wisdom file OR generate a new plan if none exists.  This wisdom will be saved for future use (see point 1) if the -W option is specified; we do not attempt to update the system-wide wisdom file.
   
4.  FFTW already has a notion of a default system-wide wisdom file.  If the -W option is not specified we attempt to load it; if it does not exist we don't complain.
   
Some notes:

- specifying  wisdom file on the Raspberry Pi *greatly* reduces the startup time on subsequent runs.  Without a wisdom file, a 10K bin size with the default "measure" plan on a RPI4 takes about 30 seconds to start executing.  With a wisdom file the startup time is negligible

- FFTW's "import system wide wisdom file" seems to always fail on the RPI (and possibly other systems as well), and the 'canonical' wisdom file generated by fftwf-wisdom doesn't seem to have any hints that help anyway.  If a user wants a system-wide wisdom file that works with hackrf_sweep, it's recommended that they generate one using the -W option to hackrf_sweep and then copy it somewhere that all user accounts can access it and chmod it/set the ACL appropriately.  Determining the right way to do this on every possible OS that hackrf can be used with seems like it's way beyond the scope of this tool.

### DIFF
--- a/host/hackrf-tools/src/hackrf_sweep.c
+++ b/host/hackrf-tools/src/hackrf_sweep.c
@@ -413,25 +413,29 @@ void sigint_callback_handler(int signum)
 }
 #endif
 
-int import_wisdom(const char* path) {
+int import_wisdom(const char* path)
+{
 	// Returns nonzero
-	if (! fftwf_import_wisdom_from_filename(path)) {
-		fprintf(stderr,"Wisdom file %s not found; will attempt to create it\n",path);
+	if (!fftwf_import_wisdom_from_filename(path)) {
+		fprintf(stderr,
+			"Wisdom file %s not found; will attempt to create it\n",
+			path);
 		return 0;
 	}
 
 	return 1;
 }
 
-int import_system_wisdom() {
-	// Returns nonzero on success
+int import_default_wisdom()
+{
 	return fftwf_import_system_wisdom();
 }
 
-int export_wisdom(const char* path) {
+int export_wisdom(const char* path)
+{
 	if (path != NULL) {
 		if (!fftwf_export_wisdom_to_filename(path)) {
-			fprintf(stderr,"Could not write FFTW wisdom file to %s",path);
+			fprintf(stderr, "Could not write FFTW wisdom file to %s", path);
 			return 0;
 		}
 	}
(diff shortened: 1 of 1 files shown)


ai6aj:
No idea what the firmware failures are all about as I haven't touched that code at all

martinling:
This looks great! I've asked the folks from #1260 to test if it fixes the failure on the Pi Zero 2 W.

The build failures aren't your fault. It looks like the latest `arm-none-eabi-gcc` package in Homebrew is failing to build libopencm3 due to a missing header. It's happening on the master branch too.

martinling (approved):
This looks great, and is now confirmed to fix #1260.

mossmann:
Thank you so much for the contribution!

PR #1383, opened 2023-12-01
Add '#include <io.h>'

gvanem:
Compiling with clang-cl, gives this error:
```c
hackrf-tools/src/hackrf_sweep.c(685,3): error: call to undeclared function '_setmode'; ISO C99 and later do not support implicit function
      declarations [-Wimplicit-function-declaration]
  685 |                 _setmode(_fileno(stdout), _O_BINARY);
      |                 ^
```

fixed by simply adding `#include <io.h>`.


gvanem:
*PS*: compiling with `cl` simply gives this warning `warning C4013: '_setmode' undefined; assuming extern returning int`

martinling:
Thanks for the fix!

Issue #1286, opened 2023-02-18
Alternative Windows host tools binary download source


dmaltsiniotis:
### What feature would you like to see and why?

It looks like the Pothos SDR folks have stopped updating their packages sometime in July 2021, which traditionally have included the HackRF host tools for Windows, and are referenced in the Windows setup docs [here](https://hackrf.readthedocs.io/en/latest/installing_hackrf_software.html#windows-binaries).

We may want to consider either
a) Finding an alternative source for Windows users to download the latest Windows host tools binaries or
b) Adding the binaries directly along side the releases here in GitHub.

There was some talk a while back (perhaps in the form of a pull request I can't remember exactly) to make the artifacts generated by the GitHub Actions pipeline available for download (for example windows binaries of the host tools.) This would solve the problem as the the build pipelines are already doing the hard work of creating the Windows binaries, we would just need to add some plumbing to be able to make them accessible to users.

straithe:
Thank you for the suggestion!

martinling:
Yes, we need an alternative solution for Windows binaries if PothosSDR isn't going to continue - this is something I've raised internally as well.

What makes this difficult is that just having a build of libhackrf and the host tools on its own is not very helpful. You also need the rest of the software available - GNU Radio, Soapy/Osmo, and all the applications. And it all needs to be built using the right version of MSVC, and installed such that it uses the right version of libhackrf. Windows doesn't have a solution for managing binary dependencies in separate packages, so it all has to be bundled together.

PothosSDR has been exactly what's needed, and I think the best solution would be for that project to be continued in some form. There is an [open issue](https://github.com/pothosware/PothosSDR/issues/97) from November asking about future updates, but no response so far.

The build scripts are all in the PothosSDR repository, so in theory there is nothing stopping someone from forking it, updating some of the packages it uses and setting up automated builds using GitHub Actions. But maintaining the whole project going forward is a big job.

dmaltsiniotis:
[...]

PR #1391, opened 2023-12-18
Auto-package windows binaries on build (and fix a few bugs and warnings)

dmaltsiniotis:
This PR adds a new windows build package artifact to each build, and also addresses a number of windows-related build issues.

The primary change with this pull request is the addition of a new artifact on build called `hackrf-tools-windows.zip`. This artifact contains the combination of the windows libhackrf build and hackrf-tools build together. The intent of this artifact is to provide a convenient way for Windows users who prefer to download pre-built Windows binaries to do so from an official source. This is the first step in a longer-term vision to be able to auto-package the HackRF One libraries and tools around a more comprehensive installer.

Specific changes in this pull request are:

* Add artifact publish step for Windows libraries and host tools.
* Replace runner.workspace with github.workspace to fix inconsistent pathing issues on the build server.
* Update `checkout` and `uploadartifact` tasks to v4 to suppress warnings since v3 uses an end-of-life node version 16.
* Update CMAKE for libhackrf and hackrf-tools to also install DLLs on WIN32 platforms.
* Add a new windows-only environment variable for the vcpkg msbuild cmake toolchain makefile.

Comments and criticisms are welcome.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,6 +15,7 @@ env:
   WIN_FFTW_LIB: -DFFTW_LIBRARIES=C:/vcpkg/installed/x64-windows/lib/fftw3f.lib
   WIN_PTHREAD_INC: -DTHREADS_PTHREADS_INCLUDE_DIR=C:/vcpkg/installed/x64-windows/include
   WIN_PTHREAD_LIB: -DTHREADS_PTHREADS_WIN32_LIBRARY=C:/vcpkg/installed/x64-windows/lib/pthreadvc3.lib
+  WIN_CMAKE_TOOLCHAIN: -DCMAKE_TOOLCHAIN_FILE=C:/vcpkg/scripts/buildsystems/vcpkg.cmake
 
 jobs:
   host:
@@ -27,7 +28,7 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
 
     - name: Install dependencies (macOS)
       run: brew install fftw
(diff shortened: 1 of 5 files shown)


martinling:
@dmaltsiniotis what are the remaining issues with this? It seems like it's working?

dmaltsiniotis:
> @dmaltsiniotis what are the remaining issues with this? It seems like it's working?

Hi @martinling,

Thanks for reminding me on this PR. The last issue that I ran into with this PR is that there is a discrepancy between running CMAKE locally on Windows, vs running CMAKE on on the Windows build agents in GitHub. The difference is that when I run the CMAKE install locally, the 3rd party dependencies installed by vstx (libusb, pthreads, fftw) are copied over into the install directory. However when cmake install runs in GitHub, the three libraries are NOT copied over to the install directory, resulting in an incomplete zip package.

I have not been successful in coaxing CMAKE to force copy the DLLs over every time, but my knowledge if CMAKE is rather rudimentary.

That is the last remaining issue, otherwise the builds and the new packaging step all run successfully.

martinling:
Thanks for clarifying! What an annoying little problem to be left with.

I've had some success at using [action-tmate](https://github.com/mxschmitt/action-tmate) to get a shell on the GitHub CI runners and debug things interactively that way, might be worth trying here.

But sometimes the result is just "it behaves differently on the runner and I can't tell why!".

martinling:
@dmaltsiniotis I've just had a go at building the `release` branch (shortly to be 2024.02.1) on Windows 10 following the same steps as we use in CI, using:

- `vcpkg` at commit `62aa4492`
- CMake 3.28.3
- Visual Studio Community 2022 (minimal install with just MSVC v143, MSBuild, C runtime and W10 SDK)

I get the same result you see: everything is built correctly but the external DLL dependencies aren't copied to the `${CMAKE_INSTALL_PREFIX}/bin` directory.

dmaltsiniotis:
@martinling, I think I've finally addressed the missing dll mystery. This PR is ready to review now.

The I haven't found the _exact_ cause of the problem, but the issue is that the vcpkg makefile is not being found/invoked on the build server, but it is on my local instance. The solution was to force/hint CMAKE to find it by specifying vcpkg as an additional tool-chain. Alternatively, we could include the vcpkg cmake file as part of an `include()` in the `CMakeLists.txt`

martinling (approved):
Great!

I agree with specifying the `CMAKE_TOOLCHAIN_FILE` on the command line in the CI script, since it's specific to where that file is installed on the build server. I wouldn't want to hardcode that in the `CMakeLists.txt`.

At some point I would like to see the package-finding code in our CMake setup get smart enough to find the vcpkg-installed dependencies without having to give all the paths explicitly, but that's out of the scope of this PR.

Thanks for all your work on this, it's very much appreciated.

mossmann:
Fixes #1286

PR #1398, opened 2024-01-08
sgpio debug ci test

grvvy:
This is expected to fail on latest and pass before fefa4f0e4592b8fb235d38e60fb5daea4974ff95


martinling (approved):
Looks good to me. Test is passing as expected now that #1412 is merged.

It fails on the RTD build because this branch predates #1406, but that failure should go away when this is merged to master.

PR #1402, opened 2024-01-15
Add HTTPS transport to Dockerfile

smburdick:
`docker build` fails during `apt-get install` on macOS (Apple Silicon). Adding transport support fixes the build.


mossmann:
What version of `apt` is this for? It looks like the `apt-transport-https` package should only be needed for `apt` versions older than 1.5.

straithe:
@smburdick Checking in as we haven't had a response in a while. Can you please confirm what version of ```apt``` this is for?

Issue #831, opened 2021-02-16
Compatibility with CMake < 2.8.12 will be removed from a future version of CMake.


dzbyr:
I don't understand what I should do.

I want to apply readme file under `hackrf-2018.01.1/firmware`
I installed below packages. I think those are latest
```
arm-none-eabi-binutils-2.35.1-1  
arm-none-eabi-gcc-10.2.0-1
```
However, I have problems with `cmake`

```make
cmake ..
CMake Deprecation Warning at CMakeLists.txt:21 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


CMake Deprecation Warning at /usr/share/cmake-3.19/Modules/CMakeForceCompiler.cmake:75 (message):
  The CMAKE_FORCE_C_COMPILER macro is deprecated.  Instead just set
  CMAKE_C_COMPILER and allow CMake to identify the compiler.
Call Stack (most recent call first):
  /home/axe/Downloads/hackrf-2018.01.1/firmware/toolchain-arm-cortex-m.cmake:27 (CMAKE_FORCE_C_COMPILER)
  build/CMakeFiles/3.19.4/CMakeSystem.cmake:6 (include)
  CMakeLists.txt:24 (project)


CMake Deprecation Warning at /usr/share/cmake-3.19/Modules/CMakeForceCompiler.cmake:89 (message):
  The CMAKE_FORCE_CXX_COMPILER macro is deprecated.  Instead just set
  CMAKE_CXX_COMPILER and allow CMake to identify the compiler.
Call Stack (most recent call first):
  /home/axe/Downloads/hackrf-2018.01.1/firmware/toolchain-arm-cortex-m.cmake:28 (CMAKE_FORCE_CXX_COMPILER)
  build/CMakeFiles/3.19.4/CMakeSystem.cmake:6 (include)
  CMakeLists.txt:24 (project)


-- Cross-compiling with the gcc-arm-embedded toolchain
-- Toolchain prefix: /usr/arm-none-eabi
-- The ASM compiler identification is GNU
-- Found assembler: /usr/bin/arm-none-eabi-gcc
-- dfu-suffix not found: not building DFU file
-- Configuring done
-- Generating done
-- Build files have been written to: /home/axe/Downloads/hackrf-2018.01.1/firmware/hackrf_usb/build

```
[...]

PR #1403, opened 2024-01-16
Update CMAKE minimum version to 2.8.12

smburdick:
To remove the CMAKE build [deprecation warning](https://github.com/greatscottgadgets/hackrf/issues/831), update the minimum required version.

### DIFF
--- a/host/CMakeLists.txt
+++ b/host/CMakeLists.txt
@@ -1,6 +1,6 @@
 #top dir cmake project for libhackrf + tools
 
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.1.3)
 project (HackRF C)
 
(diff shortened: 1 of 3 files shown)


miek:
Could this be updated to 2.8.12 as the warning suggests, or is there a reason why it should be updated to a higher version?

smburdick:
> Could this be updated to 2.8.12 as the warning suggests, or is there a reason why it should be updated to a higher version?

It could. I had no reason to make it that specific version. I can retry with 2.8.12.

martinling:
The minimum supported version has been bumped again, so we now get:

```
 CMake Deprecation Warning at libhackrf/CMakeLists.txt:24 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of CMake.
```

I guess we can probably bump to 3.5 at this point (released 2016).

mossmann:
I think 2.8.12 is good enough for today's release. Thank you for the contribution!

PR #1409, opened 2024-01-30
Added ccache speedup compilation

GermanAizek:
More info: https://ccache.dev


mossmann:
We think this may cause more trouble than it is worth. No many folks compile firmware. Those who can already enable ccache with an environment variable.

PR #1411, opened 2024-02-06
HackRF One r10

mossmann:
HackRF One r10 is based on r8, reverting most of the changes made in r9.

Major differences between r8 and r10:

- add reverse current protection diode to antenna port power supply
- change PCB stack-up
- update design to KiCad 6


catalinalb:
Hi, could you replace the micro usb connector with USB C? 
or could you share the KiCad Files in order to change it myself? 

I just fried my original hacrkf board last night, The NXP Chip was getting to 108 degrees C. 
So I measured around and on VDD I was getting only 2.5v instead of 3.3.
Connected a 18650 cell there while tring to communicate over usb cable (with 5v removed from the usb cable) but no success.
No I get only one LED that lights dim and the chip is still getting to 100 degrees :( 

Thank you,
Catalin Alb

mossmann:
> Hi, could you replace the micro usb connector with USB C?
> 
> I just fried my original hacrkf board last night, The NXP Chip was getting to 108 degrees C. So I measured around and on VDD I was getting only 2.5v instead of 3.3. Connected a 18650 cell there while tring to communicate over usb cable (with 5v removed from the usb cable) but no success. No I get only one LED that lights dim and the chip is still getting to 100 degrees :(
> 
> Thank you, Catalin Alb

HackRF One r10 is already in production. Further modifications to the design will not be made. If you need technical support, please open a separate issue or join the #hackrf channel on the Great Scott Gadgets Discord server.

miek (approved):

PR #1483, opened 2024-09-21
Implement RAW_IO with WinUSB

JVital2013:
At the request of @martinling, here's a draft PR that implements RAW_IO on Windows, using libusb/libusb#1512. The change checks if RAW_IO is available, and if it is, it turns it on for the `RX_ENDPOINT_ADDRESS`. I've made it silently ignore the result of `libusb_endpoint_set_raw_io`, since in the unlikely event of the call not working, I'd think we should continue on anyway.

This change fixes sample drops on problematic Windows machines. I've marked the PR a draft because this change will require `#if LIBUSB_API_VERSION >= 0x0100010B ... #endif` to prevent issues when building with older versions of LibUSB. The problem is:

- The change isn't actually merged into LibUSB yet (hopefully it is soon!)
- `LIBUSB_API_VERSION` has not been bumped yet. It's still sitting at 0x0100010A, so there's no good way to detect the presence of this function

While we wait for upstream devs, let me know if you'd like to see any changes to this!

### DIFF
--- a/host/libhackrf/src/hackrf.c
+++ b/host/libhackrf/src/hackrf.c
@@ -728,6 +728,13 @@ static int hackrf_open_setup(libusb_device_handle* usb_device, hackrf_device** d
 		return HACKRF_ERROR_NO_MEM;
 	}
 
+#if LIBUSB_API_VERSION >= 0x0100010C
+	// WinUSB: Use RAW_IO to improve throughput on RX
+	if (libusb_endpoint_supports_raw_io(usb_device, RX_ENDPOINT_ADDRESS) == 1) {
+		libusb_endpoint_set_raw_io(usb_device, RX_ENDPOINT_ADDRESS, 1);
+	}
+#endif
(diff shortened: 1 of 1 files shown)


martinling:
Yes, that looks fine - after all, I designed that libusb API specifically so that it could be this simple! :grin:

You'll need a space after the `if` to satisfy the `clang-format` check. You can run `tools/reformat-source.sh` to reformat automatically.

Other than that we just need to wait for the official `LIBUSB_API_VERSION` bump.

JVital2013:
The RAW_IO API has been merged upstream, and support can now be clearly identified by `#if LIBUSB_API_VERSION >= 0x0100010C ... #endif`. This PR should be good to go now

martinling (approved):

martinling:
Thank you! Very glad to finally have this done - hope to see the libusb 1.0.30 release soon.

PR #1484, opened 2024-09-26
Report whether other devices are sharing the USB bus

martinling:
Adds a check in `hackrf_info` for whether the device is on a dedicated bus.


mossmann:
Is there a reason `hackrf_device_list_bus_sharing()` must be in libhackrf, or could it be in `hackrf_info`?

martinling:
I started off with all the code in `hackrf_info` - you can see that version in [this commit](https://github.com/greatscottgadgets/hackrf/commit/f4612c34d89b8055375d65d93c0bdfc23aeefadb).

I changed it for two reasons:

1. Because the code needs to call `libusb_get_bus_number` and `libusb_get_parent`, it becomes necessary for `hackrf_info.c` to include `libusb.h`, and also to build the binary with the necessary linker flags to link it to `libusb` directly rather than just transitively via `libhackrf`. That would have required some changes in the build system, and I suspected that it might also cause some issues for distros.

2. I figured this might be useful information for GUI tools as well, so having a library function that did the necessary iterating and counting seemed like a nice idea.

mossmann (approved):
Very nice, thank you!

PR #1493, opened 2024-10-18
Handle CTRL+BREAK signals hackrf-tools on Windows

scootermon:
When running `hackrf_transfer` as a subprocess under Windows we've found that it's exceptionally hard to stop the process.
It turns out to be much easier to send a `CTRL_BREAK_EVENT` signal to a process than a `CTRL_C_EVENT`.

### DIFF
--- a/host/hackrf-tools/src/hackrf_sweep.c
+++ b/host/hackrf-tools/src/hackrf_sweep.c
@@ -415,7 +415,7 @@ static hackrf_device* device = NULL;
 #ifdef _MSC_VER
 BOOL WINAPI sighandler(int signum)
 {
-	if (CTRL_C_EVENT == signum) {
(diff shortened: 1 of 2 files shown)


martinling:
The change looks fine to me, but I'm curious why it's easier to send one than the other.

martinling (approved):
Thanks for the clarification.

I wondered if the unreliability might be something about the different threads involved, and which one ended up getting the event - we had some issues before with signal masks on worker threads spawned by libhackrf.

But from some reading it seems that on Windows, a console event causes a new thread to be started to handle that event, rather than the Unix model where the signal is received by an existing thread.

In any case, it seems perfectly reasonable that we should stop on a Ctrl-Break event in the same way as we do for a Ctrl-C event, so I'm fine with making this change.

mossmann:
Thank you for the contribution!

PR #1512, opened 2024-12-01
Add AGM CPLD to portapack detection routine

bernd-herzog:
This pull request adds the AGM Microelectronics AG256SL100 to the portapack detection routine.

Almost all new portapacks are now made with the AGM CPLD. Without the ability to detect them in the stock hackrf firmware, users will just see a black screen and might think there is a defect as long as they didn't replace the firmware. So this should reduce user confusion and might even reduce support requests on both sides.

In the portapack firmware this fix is already injected into the "hackrf mode", but our compilation is somewhat messy now (can't just use your CMakeLists.txt).

### DIFF
--- a/firmware/common/portapack.c
+++ b/firmware/common/portapack.c
@@ -614,7 +614,11 @@ static uint32_t jtag_pp_idcode(void)
 
 static bool portapack_detect(void)
 {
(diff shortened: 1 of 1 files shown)


mossmann (approved):
Wonderful! Thank you for the contribution!

PR #1517, opened 2024-12-16
Add a shim header to support locking without ldrex/strex

martinling:
Required to allow the USB queue code to be built for the M0.

### DIFF
--- a/firmware/common/locking.h
+++ b/firmware/common/locking.h
@@ -26,6 +26,12 @@
 
 #include <libopencm3/cm3/sync.h>
 
+/* Primitives for implementing locking.
+ *
+ * Must always be used in a pair, with a call to load_exclusive being
+ * followed immediately or near-immediately by a call to store_exclusive().
+ * Failure to observe this rule may lead to undefined results. */
+
 // Use ldrex and strex directly if available.
 // Otherwise, disable interrupts to ensure exclusivity.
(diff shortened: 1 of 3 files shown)


miek (commented):
One minor concern I have is about the side-effects of using one of the functions on their own (and leaving interrupts disabled for example) but, given that ldrex/strex must be used in a pair, I'm ok with assuming that for these functions too. I think it'd be worth adding a comment to that effect though.

martinling:
Both changes made.

miek (approved):

gullradriel:
Thanks a lot :-)

gullradriel:
Back from testing.
Sorry for pinging you anew, @martinling .
Compiling on M0 is giving an error for undefined references to the intrinsic __disable_irq and __enable_irq.
The only two solutions for us were to either #include "hal.h" from chibios, or use the asm alternatives to those functions.
Since I don't think hackrf code is using chibios, would it be possible to use the following code instead:
```
static inline uint32_t load_exclusive(volatile uint32_t* addr)
{
	__asm volatile ("cpsie i" ::: "memory"); 
   	return *addr;
}

static inline uint32_t store_exclusive(uint32_t val, volatile uint32_t* addr)
{
	*addr = val;
	__asm volatile ("cpsid i" ::: "memory"); 
	return 0;
}
```

PR #1519, opened 2024-12-17
define replacement for enable and disable irq for m0

gullradriel:
Following the comment here: https://github.com/greatscottgadgets/hackrf/pull/1517#issuecomment-2549482791

We are getting undefined reference to enable/disable irq calls.

I'm providing a fix for build on m0, based on the definition that can be found in chibios.

### DIFF
--- a/firmware/common/locking.h
+++ b/firmware/common/locking.h
@@ -40,14 +40,14 @@
 #else
 static inline uint32_t load_exclusive(volatile uint32_t* addr)
 {
-	__disable_irq();
+	__asm volatile("cpsid i");
 	return *addr;
 }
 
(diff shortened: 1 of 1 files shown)


miek:
Sounds good to me.

I think it would be worth adding the `"memory"` clobber you mentioned in #1517 though. As far as I can tell from the docs & various discussions online, the compiler can choose to re-order instructions potentially putting the load/store on the wrong side of the asm statement. Adding that clobber argument acts as a memory barrier and prevents that from happening.

gullradriel:
Thanks, I'm adding it right now.

miek (approved):

gullradriel:
Thanks :-)

PR #1551, opened 2025-05-19
Update section about windows binaries

philippuhl:
The most current PothosSDR binary in the link is from 2021, while the build artifacts always provide the most current binary.
They also are available directly from this repository this way, instead of relying on a third party.


philippuhl:
I added the Information about radioconda. I would still suggest leaving the hint about the build artifacts in the documentation, as this is currently the only source of the binaries that don't require any third party.

PR #1555, opened 2025-05-23
Replace fftSize with more meaningful name for num_fft_bins

albayenes:
`fftSize` misleads while reading the code. `num_fft_bins` is more suitable than `fftSize`.


mossmann:
Thank you!

PR #1558, opened 2025-06-06
ci: improve sgpio-debug test reliability and error messaging

grvvy:


martinling (changes requested):
I think a slightly different approach is needed here.

It should not be possible for the _firmware_ to produce non-sequential bytes when compiled in the SGPIO debug mode. If that happens, it's a serious problem and it's important that we know about it, rather than brushing an intermittent failure under the carpet because the test was simply retried.

However, it's possible that by the time the data arrives at the _host_, some chunk of it has been dropped because the firmware ran out of buffer space and had to drop samples. In that sense, it's to be expected that sometimes the test may fail, if the USB bus isn't running smoothly enough.

The good thing is, we can distinguish these two cases. Running `hackrf_debug --state` after the test will report the number of shortfalls (i.e. buffer underruns) during the last TX/RX operation.

So the logic should be:

1. If the shortfall count is zero, the test is valid. If the data was fully sequential, the test passes. If the data was not fully consecutive, there is a firmware bug and the test fails.
2. If the shortfall count is not zero, the test is invalid, and may be retried.

Issue #1199, opened 2022-10-14
cmake dev warning FindUSB1.cmake -- find_package(USB1 REQUIRED)


mpouyann:
### What type of issue is this?

permanent - occurring repeatedly

### What issue are you facing?

while compiling libhackrf cmake command is raising a dev warning.

I did made some changes in me setup to fix this issue, but I'm not sure if a larger check / modification should be made.
Fix modifications done :
     modified /host/libhackrf/CMakeLists.txt (48)
                  -- find_package(USB1 REQUIRED )
                  ++ find_package(LIBUSB REQUIRED )

       renamed /host/cmake/modules
                  --  FindUSB1.cmake
                  ++  FindLIBUSB.cmake

### What are the steps to reproduce this?

git clone hackrf
cd host
mkdir build
cd build
cmake ..

dev warning is raised

### Can you provide any logs? (output, errors, etc.)

CMake Warning (dev) at /usr/share/cmake-3.24/Modules/FindPackageHandleStandardArgs.cmake:438 (message):
  The package name passed to `find_package_handle_standard_args` (LIBUSB)                                    
  does not match the name of the calling package (USB1).  This can lead to                                   
  problems in calling code that expects `find_package` result variables                                      
  (e.g., `_FOUND`) to follow a certain pattern.                                                              
Call Stack (most recent call first):                                                                         
  cmake/modules/FindUSB1.cmake:39 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)                                        
  libhackrf/CMakeLists.txt:48 (find_package)                                                                 
This warning is for project developers.  Use -Wno-dev to suppress it.

maitbot:
Fixed in my fork which refactors the CMake build system.
https://github.com/maitbot/hackrf/tree/mait/update-cmake

martinling:
Fixed in #1586.

Issue #1193, opened 2022-09-29
[Bug Report]: static libs are unconditionally built/installed on linux


ZeroChaos-:
### What type of issue is this?

permanent - occurring repeatedly

### What issue are you facing?

static libs are installed by not needed or requested

### What are the steps to reproduce this?

cmake
make
make install

### Can you provide any logs? (output, errors, etc.)

https://bugs.gentoo.org/724906

mossmann:
We provide multiple options for building libhackrf.

- only dynamic libraries: `make hackrf`
- only static libraries: `make hackrf-static`
- both: `make all` or `make`

Are these options insufficient in some way?

mossmann:
Oh, it looks like `make install` installs both static and dynamic libraries even if you do `make hackrf-static install`. Would it be acceptable to add a separate `make install-static` target and change the `make install` behavior to only build dynamic?

ZeroChaos-:
in Gentoo libhackrf and hackrf-tools are currently divided into two separate packages.  It seems that building this way doesn't support "make hackrf"
```
>>> Compiling source in /var/tmp/portage/net-libs/libhackrf-2022.09.1/work/hackrf-2022.09.1/host/libhackrf ...
make -j12 -l12 hackrf 
make: *** No rule to make target 'hackrf'.  Stop.
```
regular `make` (which I assume runs `make all`) works fine, just builds more than I want.
I am not great at reading the CMakeFiles, so I can't make much of a suggestion here.

mossmann:
I'm not sure how you are getting `No rule to make target 'hackrf'`. It looks like your output is truncated. Can you provide more context for this error?

This works for me:
```
cd host/libhackrf/
mkdir build
cd build
cmake ..
make hackrf
```

ZeroChaos-:
```
>>> Emerging (1 of 1) net-libs/libhackrf-2022.09.1::maintainer-zero
 * hackrf-2022.09.1.tar.xz BLAKE2B SHA512 size ;-) ...                                                                                                                [ ok ]
>>> Unpacking source...
>>> Unpacking hackrf-2022.09.1.tar.xz to /var/tmp/portage/net-libs/libhackrf-2022.09.1/work
>>> Source unpacked in /var/tmp/portage/net-libs/libhackrf-2022.09.1/work
>>> Preparing source in /var/tmp/portage/net-libs/libhackrf-2022.09.1/work/hackrf-2022.09.1/host/libhackrf ...
[...]

PR #1578, opened 2025-08-17
Mait/modern cmake

maitbot:
# modern cmake
Refactor host CMake build.

Replace pull request 1016
"Host software CMake cleanup: attempt #2 #1016"
https://github.com/greatscottgadgets/hackrf/pull/1016
to adress more issues.

Update cmake_minimum_required.
(GNU Radio picked 3.16, because Ubuntu 20.04 used 3.16.3, and that is
now more than 5 years ago.)
Like pull request 1514
"updated minimum required cmake version to 3.16 #1514"
https://github.com/greatscottgadgets/hackrf/pull/1514
but just for the host (not firmware) build.

Update project() call.
CMake project handles VERSION now too.
Defines PROJECT_VERSION and PROJECT_VERSION_MAJOR etc., which will be used
to set CMake library target properties for VERSION and SOVERSION.

Add some overall build defaults,
set(CMAKE_VERBOSE_MAKEFILE ON)
set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
These settings help to not hide details from the build logs, and
editors using LSP will want to consume the generated compile_commands.json file.
(vim can use LSP with clangd, and emacs has LSP and eglot modes.)

Make use of CMake's GNUInstallDirs.
Smart way to set install DESTINATIONs.
Update all DESTINATION to use GNUInstallDirs variables.
(except MSVC builds - someone else want to change this?)

Just use CMake's FindThreads.
Use the Threads::Threads CMake target.
Remove cmake/modules/FindThreads.cmake
On Windows, use find_package(PThreads4W) and the
resulting PThreads4W::PThreads4W also used by the
Windows USB library.

Rename FindUSB1 to FindLIBUSB, and have it provide the
LIBUSB::LIBUSB CMake target. (This also avoids warnings about
confusion between LIBUSB vs USB1 definitions.) Solving
"cmake dev warning FindUSB1.cmake -- find_package(USB1 REQUIRED) #1199"
https://github.com/greatscottgadgets/hackrf/issues/1199

# libhackrf:
Remove c_sources and c_headers CMake variables, they just add a
potentially confusing layer of indirection. Just be explicit when
referencing source files.

See CMake documentation for cmake-packages(7)
Install CMake Config-file package files for users.
Maybe include(GenerateExportHeader)? Added in version 3.12: Added
support for C projects. Previous versions supported C++ project only.
Solves issue "Install project config files for cmake projects to use libhackrf #391"
https://github.com/greatscottgadgets/hackrf/issues/391

# hackrf-tools:
Use FindFFTW3.cmake from GNU Radio instead of FindFFTW and use the
modern CMake target fftw3f::fftw3f it provides.

Build with Modern CMake targets.

# CMake options:
Keep option INSTALL_UDEV_RULES, "Install udev rules for the HackRF".

Add options to allow shared or static library to be disabled,
ENABLE_STATIC_LIB
ENABLE_SHARED_LIB.
Solves "[Bug Report]: static libs are unconditionally built/installed on linux #1193"
https://github.com/greatscottgadgets/hackrf/issues/1193

Add option to disable hackrf_sweep - which allows building without FFT library.
option(ENABLE_HACKRF_SWEEP
Solves issue "Building hackrf host code w/out fftw #729"
https://github.com/greatscottgadgets/hackrf/issues/729

Add option to control a DISABLE_USB_DEVICE_DISCOVERY compile definition,
which could be used to alter libusb usage as needed on Android.
+option(DISABLE_USB_DEVICE_DISCOVERY
Like the CMake portion of pull request 1150
"Non-root android support #1150"
https://github.com/greatscottgadgets/hackrf/pull/1150

# update scripts
Also add a commit for a more modern use of CMake in a script:
cmake -B builddir
cmake --build builddir

This commit is only CMake build system changes. No modified C code.
Checked to ensure project build with both Makefile and Ninja generators.
CI builds OK, including  Windows build.


martinling (commented):
Hi @maitbot,

I'm working on reviewing this and have a few questions & notes:

1. It's a bit awkward having most of this in one commit, with multiple unrelated changes including a lot of reformatting. Was this by any chance squashed from a longer branch, and if so, would it be possible to see that branch - even if it's messy? I'd prefer to have these changes broken down a bit more, and I'm happy to help with doing that, but it would be easier if I had more history.

2. You specify that this is intended to replace #1016. In your view, are there any benefits of that earlier PR which are not included in this one, or do you see it as a complete replacement at this point?

3. Since you posted this, we've bumped our CMake minimum version to 3.10 rather than 3.16, for the reasons discussed in https://github.com/greatscottgadgets/hackrf/pull/1514#issuecomment-3310128758. We've removed the obsolete Appveyor build that was failing on this PR, and I've added matrix testing across CMake versions to Github Actions.

4. I have a rebase of your PR at #1586 with some changes that I'd appreciate your comments on - these were primarily to keep compatibility to CMake 3.10, since that turned out to be fairly straightforward to do, but I also spotted one bug:

    1. Stick to `project(HackRF C)` rather than using the longer signature including `HOMEPAGE_URL` (added in 3.12). I don't think we gain anything from adding the URL there.
    2. Avoid the use of `GenerateExportHeader`, which only supported C++ projects until 3.12. Unless I'm missing something, this wasn't actually being used, since the line installing `hackrf_export.h` was commented out.
    3. Add an `ARCHIVE DESTINATION` argument to the Unix `install` command in `libhackrf_common_settings` function (needed for both 3.10 and 3.12, at least).
    4. Correct an error spotted in `FindFFTW3f.cmake`:  the `HINTS` for `FFTW3f_INCLUDE_DIRS` should include `PC_FFTW3f_INCLUDE_DIRS` (not `DIR`), as per [FindPkgConfig](https://cmake.org/cmake/help/latest/module/FindPkgConfig.html#command:pkg_check_modules) docs. This broke some builds on macOS with Homebrew, where each package gets its own prefix so the hint from `pkg-config` is essential.

martinling:
Thanks for the answers. Let's move forward with #1586 then - I'd like to make some further edits to it to split up the changes, and incorporate your notes from above about what has changed and which issues that addresses.

PR #1586, opened 2025-09-19
Modern CMake, rebased

martinling:
Rebase of #1578.

PR #1589, opened 2025-10-01
Fix missing shared library version

atsampson:
The libtarget variable isn't defined here (it's used in a helper function above). This caused libhackrf to be built and installed without a version number.

This fixes a change from #1586.


mossmann:
Thank you for the fix!

PR #1597, opened 2025-10-05
CI: Add builds on Windows using MSYS2

martinling:
This PR adds a firmware build on Windows, and an additional host software build on Windows, built using gcc under MSYS2 with the UCRT64 environment.

Uses the [`setup-msys2`](https://github.com/msys2/setup-msys2) action to setup MSYS2 on Windows runners and install required packages.


grvvy:
Looks good to me!

Issue #1363, opened 2023-10-12
Periodic small glitch in the received signal


VioletGiraffe:
### What type of issue is this?

permanent - occurring repeatedly

### What issue are you facing?

The issue looks like this in our software, this is just a visualization of the raw I/Q values, 4 pixels per IQ sample:

![unnamed](https://github.com/greatscottgadgets/hackrf/assets/1175926/1266a698-ebc0-402a-8eeb-aaa2b516aef0)
This was confirmed with two different HackRF units, same behavior.

To verify, I have connected a third unit, bought from a different seller, to a different computer, and used SDR++ to record the signal. This is different frequency, too - 226 MHz, I just scrolled around until I found a strong carrier signal with no apparent modulation. I recorded raw data and opened it as audio in Sony Sound Forge, and here it is:

![image](https://github.com/greatscottgadgets/hackrf/assets/1175926/124a174d-d482-4fa0-b1be-1eec87707dd1)
And here it is again, around 8000 samples later. The period is slightly less than 8192, hard to calculate precisely in a sound editor, I'm just using 3rd-party software to demonstrate that our own software is not the culprit.

![image](https://github.com/greatscottgadgets/hackrf/assets/1175926/0791bb4d-6a05-43ec-afed-747de88db757)

Initially, we were developing an app that receives analog video, the glitch looks like a white dot on the left. This is at 1080 MHz carrier frequency. Each line is 1024 samples (2048 bytes), the "ruler" at the very left is 32 zeros at each 256 KB and 16 zeroes at each 16 KB block, we added them right after libusb bulk transfer returns (async transfers). Each white dot slightly off the left side is the glitch. As you can see, it is not absolutely periodic, but the pattern is there. The vertical line is a perfect straight line, so it always occurs at a perfect multiple of some magic sample count (and a multiple of 2048).

![unnamed](https://github.com/greatscottgadgets/hackrf/assets/1175926/31ce443c-d8a0-491c-b014-c702f4b2772a)

Any ideas?
All our 3 units are running the current 2023 firmware.

### What are the steps to reproduce this?

In order to reproduce, I think it is enough to tune to a strong clean (unmodulated) carrier signal so that you can see a more or less clean sine wave with decent amplitude. Plot the sample values against time, then the "dislocations" will be obvious. Oh, and this is all at 20 MHz, we have not tested lower sampling rates as we need the 20 MHz bandwidth.

martinling:
[...]

PR #1601, opened 2025-10-20
Firmware: Add an additional layer of DMA-assisted sample buffering

### LINKED ISSUE #1503: hackrf_transfer -c produces possible buffer underruns

While measuring the output of a *clone* HackRF one with a proper spectrum analyser, I noticed the power output fluctuating, and with slower sweep times, the spectrum looked exactly like an OOK spectrum.
I confirmed the result using time-domain measurements with a low-frequency carrier - the output went to 0 for extended periods (milliseconds) somewhat randomly. 
I suspected software error, and [wrote a small piece of C++ code](https://gist.github.com/Sasszem/350c24559f6b5410c9badfe4fb3cebd6) on my own to transmit. My code did NOT produce the same effect, confirming a software bug in or triggered by `hackrf_transfer`.
I am planning to investigate the real cause, but in the meantime, I'm opening this issue in case anyone else encountered the same effect.

- run `hackrf_transfer -f 10000000 -a 0 -c 127 -x 47` (10 MHz output with maximum gain settings without amplifier)
- measure with spectrum analyser, 10Meg center, 2Meg SPAN and <1khz RBW. You should (or rather should not) see 1/x type OOK spectrum. WARNING: set your reference level high enough or use an external attenuator so you will not damage your instrument!
- measure with oscilloscope, time scale SLOWER so you can see amplitude variations in the millisecond scale, like 1ms / div

_No response_

martinling:
This PR attempts to implement the buffer scheme described in https://github.com/greatscottgadgets/hackrf/issues/1363#issuecomment-1850452973.

The M0 SGPIO code continues to use the existing 32KB sample buffer, which has special placement.

An additional 32KB USB buffer is placed in `ram_local1`, above the `.text` section.

The GPDMA controller is used to transfer samples between the sample buffer and USB buffer.

The initial commits in this PR set up the new buffer scheme, whilst using `memcpy` running synchronously on the M4 core as a placeholder for the DMA operations. This works correctly up to around 8Msps.

The final commit switches to the DMA implementation.

### DIFF
--- a/firmware/common/LPC4320_M4_memory.ld
+++ b/firmware/common/LPC4320_M4_memory.ld
@@ -26,7 +26,8 @@ MEMORY
 {
 	/* rom is really the shadow region that points to SPI flash or elsewhere */
 	rom (rx)  : ORIGIN = 0x00000000, LENGTH =  1M
-	ram_local1 (rwx) : ORIGIN = 0x10000000, LENGTH =  96K
+	ram_local1 (rwx) : ORIGIN = 0x10000000, LENGTH =  64K
+	ram_usb (rw) : ORIGIN = 0x10010000, LENGTH = 32K
 	ram_local2 (rwx) : ORIGIN = 0x10080000, LENGTH =  32K
 	ram_sleep (rwx) : ORIGIN = 0x10088000, LENGTH = 8K
 }
--- a/firmware/common/LPC4330_M4_memory.ld
+++ b/firmware/common/LPC4330_M4_memory.ld
@@ -26,7 +26,8 @@ MEMORY
 {
 	/* rom is really the shadow region that points to SPI flash or elsewhere */
 	rom (rx)  : ORIGIN = 0x00000000, LENGTH =  128K
-	ram_local1 (rwx) : ORIGIN = 0x10000000, LENGTH =  128K
+	ram_local1 (rwx) : ORIGIN = 0x10000000, LENGTH =  96K
+	ram_usb(rw) : ORIGIN = 0x10018000, LENGTH = 32K
 	ram_local2 (rwx) : ORIGIN = 0x10080000, LENGTH =  64K
 	ram_sleep (rwx) : ORIGIN = 0x10090000, LENGTH = 8K
 }
--- a/firmware/common/LPC43xx_M4_memory.ld
+++ b/firmware/common/LPC43xx_M4_memory.ld
@@ -41,5 +41,6 @@ MEMORY
 lz4_in_buf = ORIGIN(ram_lz4_in);
 lz4_out_buf = ORIGIN(ram_lz4_out);
 usb_samp_buffer = ORIGIN(ram_samp);
+usb_bulk_buffer = ORIGIN(ram_usb);
 m0_state = ORIGIN(ram_shared);
 PROVIDE(__ram_m0_start__ = ORIGIN(ram_m0));
--- a/firmware/common/LPC43xx_M4_memory.ld
+++ b/firmware/common/LPC43xx_M4_memory.ld
@@ -35,11 +35,11 @@ MEMORY
 	 * to get performance benefit of having two USB buffers addressable
 	 * simultaneously (on two different buses of the AHB multilayer matrix)
 	 */
-	ram_usb (rwx)     : ORIGIN = 0x20008000, LENGTH = 32K
+	ram_samp (rwx)     : ORIGIN = 0x20008000, LENGTH = 32K
 }
 
-usb_bulk_buffer = ORIGIN(ram_usb);
 lz4_in_buf = ORIGIN(ram_lz4_in);
 lz4_out_buf = ORIGIN(ram_lz4_out);
+usb_samp_buffer = ORIGIN(ram_samp);
 m0_state = ORIGIN(ram_shared);
 PROVIDE(__ram_m0_start__ = ORIGIN(ram_m0));
--- a/firmware/common/fpga_selftest.c
+++ b/firmware/common/fpga_selftest.c
@@ -37,8 +37,8 @@
 #include "streaming.h"
 
 // USB buffer used during selftests.
-#define USB_BULK_BUFFER_SIZE 0x8000
-extern uint8_t usb_bulk_buffer[USB_BULK_BUFFER_SIZE];
+#define USB_SAMP_BUFFER_SIZE 0x8000
+extern uint8_t usb_samp_buffer[USB_SAMP_BUFFER_SIZE];
 
 static int rx_samples(const unsigned int num_samples, uint32_t max_cycles)
 {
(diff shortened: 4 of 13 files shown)


mossmann:
On my Linux test host, this improves maximum shortfall-free RX sample rate from 19.2 Msps to 21.8 Msps and improves TX from 20.1 Msps to 21.8 Msps.

mossmann (dismissed):
I haven't spotted the bug, but I'm seeing short TX with `hackrf_transfer -n` truncated by a few milliseconds. Looks like the transmissions are about 9000 samples short.

mossmann:
Transmission duration does not change linearly with the number of samples.

Testing with `hackrf_transfer -a 1 -x 31 -c 127 -f 2500000000 -n 32768 -s 2000000`

32768 samples: 8.88 ms (expected 16.38 ms)
32769 samples: 12.29 ms (expected 16.38 ms)
33000 samples: 12.41 ms (expected 16.50 ms)
40000 samples: 12.93 ms (expected 20.00 ms)
40960 samples: 13.03 ms (expected 20.48 ms)
40961 samples: 16.85 ms(expected 20.48 ms)

martinling:
I believe this is due to the current behaviour of `hackrf_enable_tx_flush`.

When that function is used, the host will send 32KB of zeroes after the end of the provided TX data, before making the control request to switch transceiver mode back to idle.

The theory behind that approach was that, since there's only 32KB of space to store samples, once you've sent that many zeroes and the device has accepted them, the earlier samples you care about must have been transmitted.

With this PR, we double the buffer space, so the host would now need to send 64KB of zeroes to achieve the same. If you change `DEVICE_BUFFER_SIZE` in `libhackrf/src/hackrf.c` to 65536, I think that will stop the behaviour you're seeing.

Unfortunately, we hardcoded that figure rather than adding a vendor request to query it from the device. My recollection is that we discussed that decision verbally at the time, but since I'd already made some attempts to increase the buffer size without success, we thought that 32KB figure wasn't ever going to change and it wasn't worth adding this query.

Our options are:

- Bump the `DEVICE_BUFFER_SIZE` in libhackrf. This is no good, because users will get short TX if running older software with newer firmware, just as you're seeing now.

- Add a vendor request to query the buffer size, defaulting to 32KB if the request is not supported. Same problem, because older software won't make the request. We would have needed to add this at the same time we added `hackrf_enable_tx_flush`.

- Make the firmware emulate the previous behaviour, as follows: Once the host has sent the mode change request to leave TX mode, stop making transfers into the USB buffer, but move all existing data from the USB buffer into the sample buffer before actually shutting down TX.

I'll implement the latter.

martinling:
Ensuring that all data in the USB buffer is moved to the sample buffer before stopping TX has fixed the non-linearity you were seeing, but I'm now seeing timing coming up short by what looks like a consistent 4.096ms, i.e. 8KB of samples or one DMA transfer. I'll see about tracking down the remaining bug.

martinling:
Some experimentation with looking at ramped signals on a scope shows that the transmission includes the first 32KB preloaded into the sample buffer; is missing the next 8KB, which is the first to be transferred by DMA; then all subsequent samples are transmitted correctly. This is true from the memcpy version onwards, regardless of whether `DMA_TRANSFER_SIZE` is 16KB or 8KB.

martinling:
Fixed that - still some issues with small sample counts.

martinling:
I've rebased this on `main` and done some further testing of the forward/back compatibility.

Everything seems to be working as expected, so this is ready for review again.

The forward/back compatibility logic is:

Host: if API version < 1.10, assume buffer is 32KB. If API version >= 1.10, ask for buffer size. The rest of the flush logic is unchanged (send enough zeroes, then assume all samples were transmitted).

Firmware: if the host didn’t ask for our buffer size, NAK any request to stop TX until all samples are transmitted.

This should result in everything working optimally, except in the case of old host software with new firmware, in which case the device takes a little longer to leave TX.

mossmann:
With this change I'm seeing RX hang when adjusting the sample rate mid-stream, for example with the flowgraph in https://github.com/greatscottgadgets/hackrf/issues/1629#issuecomment-3693083275

martinling:
This has been rebased on `main`. I can no longer reproduce the RX hang seen in https://github.com/greatscottgadgets/hackrf/pull/1601#issuecomment-3720499697, which I think was due to the concurrency problems fixed in #1648. As such I think this is ready for a fresh review.

mossmann (changes requested):
I'm still able to induce hangs with a sample rate slider. I've seen this in both TX and RX, but so far I've only induced RX failure with a PortaPack installed. The TX failure also seems easier to reproduce with a PortaPack.

I tried merging #1689, hoping it would avoid PortaPack concurrency problems, but it didn't help.

martinling:
Bummer. OK, I'll see if I can reproduce with a Portapack. In the meantime, putting this one back to draft for now.

martinling:
As discussed yesterday, I was able to reproduce the RX hang with a PortaPack attached, and examine the situation with a debugger.

It turned out the hang was caused by a race condition in the following code:

```c
volatile uint32_t dma_pending;

// ...

void transceiver_start_dma(...) {
        // ...
        gpdma_channel_enable(DMA_CHANNEL);
        dma_pending = size;
}

void dma_isr(void)
{
        gpdma_channel_disable(DMA_CHANNEL);
        GPDMA_INTTCCLEAR = (1 << DMA_CHANNEL);
        m0_state.m4_count += dma_pending;
        dma_pending = 0;
}
```

Because `transceiver_start_dma()` called `gpdma_channel_enable(DMA_CHANNEL)` before updating `dma_pending`, it was possible for an interrupt to occur between those steps. If that interrupt also took long enough to complete, the DMA transfer could complete in the background, and `dma_isr` would execute next, setting `dma_pending` to zero, only to then have it reset back to `size` by the last line of `transceiver_start_dma()`. The buffer management code would then be blocked from making further DMA transfers, as it would believe one was still pending.

The PortaPack made the bug easier to reproduce, because the USB ISR would update the baseband LPF cutoff on the LCD display. Once I realised this, I was able to reproduce the race more quickly by rapidly alternating between sample rates that caused different LPF settings to be chosen.

This race has been fixed, and the branch rewritten with a cleaner implementation and history. I'm no longer able to reproduce the RX hang and am reopening this for review.

mossmann:
I'm no longer able to reproduce the hang in RX, but I can still reproduce it quite easily in TX.

martinling:
OK, the TX hang must be a different bug then. I'll investigate.

martinling:
I was able to reproduce the TX hang, and worked out that it was due to mistakes in the calculations of `space_in_use` and `samp_buf_margin` in the TX path through `start_dma_if_possible()`.

I've fixed that, and rebased the branch to resolve new conflicts with the merge of #1701.

martinling:
Rebased on `main`.

It occurs to me we could also apply this improvement to sweep mode.

mossmann (approved):
Working great for me now! I get 21.8 Msps without shortfalls on Linux with Praline (with PortaPack), H1, and Jawbreaker in both TX and RX mode. If I hack the firmware to allow sample rates above 21.8 Msps, I see shortfalls starting at 21.9 Msps and termination of streaming at higher rates. I can't reproduce TX or RX hangs or spectrum inversion except on H1 which seems to be due to #1713, not this PR. I repeated the `hackrf_transfer -n` short transmission test and got the expected results.

mossmann:
planning to merge after #1710 and a rebase

PR #1611, opened 2025-11-26
Do not halt when FPGA configuration fails: update report instead.

mndza:


mossmann (approved):
This is great! Thank you!

PR #1612, opened 2025-11-27
New method to test RTC oscillator

martinling:
Implements a new method for testing the 32kHz RTC oscillator.

Avoids self-testing at startup due to the long startup time of the oscillator. Use `hackrf_debug -o` to run the test.

Handles these fun LPC43xx behaviours when the 32kHz oscillator is enabled, but not working due to a hardware fault:
- Attempting to write to `CREG0` to stop the oscillator will hang the MCU.
- The `MEAS` flag in the `FREQ_MON` register will never clear when attempting to measure it.

The test process is driven by the host as follows:
1. Host requests that the device enable its 32kHz oscillator.
2. Host waits 1s for oscillator startup (typical startup time: 300ms).
3. Host requests device start a frequency measurement using the CGU `FREQ_MON` register.
4. Host waits another 1ms to allow measurement to complete (needs 43us: 511 cycles at 12MHz).
5. Host requests the `FCNT` value. If the measurement has not completed (`MEAS` bit still set) the device reports zero.
6. Host checks the `FCNT` value is 1, as expected. If so, it requests that the device disable its 32kHz oscillator. If not, it does not make this request, as doing so will hang the device.

Tested on Praline r1.0.0 and HackRF One; both pass consistently, and fail consistently if the 32kHz crystal is shorted.

Also tested on rad1o, which always fails, as it has no 32kHz crystal.

### DIFF
--- a/host/hackrf-tools/src/hackrf_debug.c
+++ b/host/hackrf-tools/src/hackrf_debug.c
@@ -577,6 +577,7 @@ static void usage()
 	printf("\t-u, --ui <1/0>: enable/disable UI\n");
 	printf("\t-l, --leds <state>: configure LED state (0 for all off, 1 for default)\n");
 	printf("\t-t, --selftest: read self-test report\n");
+	printf("\t-o, --rtc-osc: test 32.768kHz RTC oscillator\n");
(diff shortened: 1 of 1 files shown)


mossmann (changes requested):
This is great, but it is missing help output for the `-o` option.

PR #1613, opened 2025-11-27
Fix zlp bug when request response length falls on a packet boundary.

antoinevg:
Previously this bug did not occur as the size of `usb_endpoint_t`'s IN transfer buffer was set to 32 bytes and control packets are 64 bytes long.

### DIFF
--- a/firmware/common/usb.c
+++ b/firmware/common/usb.c
@@ -520,7 +520,7 @@ static void copy_setup(usb_setup_t* const dst, const volatile uint8_t* const src
 	dst->length_h = src[7];
 }
 
-void usb_endpoint_init(const usb_endpoint_t* const endpoint)
+void usb_endpoint_init(const usb_endpoint_t* const endpoint, const bool enable_zlp)
 {
 	usb_endpoint_flush(endpoint);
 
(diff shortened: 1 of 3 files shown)


miek:
It looks like this change affects the bulk endpoints too? For me it seems to break TX streaming.

I'm also not sure if queueing up a whole empty transfer is the right solution - given that none of this API is packet-based, it seemed odd that there isn't a clean way to handle it. I did a bit of digging and there's a "ZLT" (zero length termination) flag that can be changed in the endpoint capabilities, which is currently disabled. It may be worth looking into enabling that on the control endpoints instead.
https://github.com/greatscottgadgets/hackrf/blob/main/firmware/common/usb.c#L540

antoinevg:
> For me it seems to break TX streaming.

Ugh of course it would. Tx for checking.

> I'm also not sure if queueing up a whole empty transfer is the right solution - given that none of this API is packet-based, it seemed odd that there isn't a clean way to handle it. I did a bit of digging and there's a "ZLT" (zero length termination) flag that can be changed in the endpoint capabilities, which is currently disabled. It may be worth looking into enabling that on the control endpoints instead. https://github.com/greatscottgadgets/hackrf/blob/main/firmware/common/usb.c#L540

Very nicely spotted indeed! I don't understand why clearing the flag fixes the issue, but it does indeed fix the issue.

I've force-pushed the branch with my changes and I also did some captures:

---

### Original

01-original

### Forcing a ZLP with a zero length transfer:

02-zero-length-transfer

### Clearing `USB_QH_CAPABILITIES_ZLT`:

03-usb-qh-capabilities-zlt

---

Tx @miek !

miek:
Nice!

> I don't understand why clearing the flag fixes the issue, but it does indeed fix the issue.

The naming on that flag is kinda backwards, setting it to 0 enables automatic ZLT (ZLP) handling and setting it to 1 disables it. It probably should've been called "ZLT_DISABLE" or something.

zlt-flag

So the USB controller is supposed to handle it, but previously we had ZLT handling disabled on all endpoints.

I guess later we might want to investigate enabling it on the OUT endpoint too. I would guess that if the host sends a transfer with a ZLP, the LPC may interpret that as two separate transfers.

----

One last little nitpick, I just noticed that the user manual under 24.9.1.1 says:
 > Endpoint capabilities and characteristics descriptor fields
 >
 > This DWord specifies static information about the endpoint, in other words, this
information does not change over the lifetime of the endpoint. **The device controller
software should not attempt to modify this information while the corresponding endpoint is
enabled.**

So it may not be safe to set the flag after calling `usb_endpoint_init`. Maybe it'd make sense to add an argument to `usb_endpoint_init` for enabling/disable ZLT and handle it in there.

antoinevg:
> I guess later we might want to investigate enabling it on the OUT endpoint too. I would guess that if the host sends a transfer with a ZLP, the LPC may interpret that as two separate transfers.

I did give it a quick try on the OUT endpoint but that breaks `hackrf_spiflash`:

04-spiflash-fail

Definitely worth looking into at a later point though, we may be wanting to do some large Control OUT transfers in the near future.

> So it may not be safe to set the flag after calling `usb_endpoint_init`. Maybe it'd make sense to add an argument to `usb_endpoint_init` for enabling/disable ZLT and handle it in there.

Oh most excellent catch! I've updated the PR with your suggestion. Tx @miek!

martinling:
It turns out this change broke `hackrf_spiflash` when *reading* flash, but nobody noticed since that's rarely used.

Before, at commit 90e15ec85f921f94c6996d8d5f3b295e05a7d7bb:
image

At commit db73ecbf5b3c201278905216a320ad156d07da1b:

image

PR #1614, opened 2025-11-28
Add register definitions for the standard FPGA bitstream.

antoinevg:


host/hackrf-tools/src/hackrf_debug.c
```diff
@@ -417,35 +417,49 @@ int rffc5072_write_register(
...
 	return result;
 }
 
-int fpga_spi_write_register(
+int fpga_read_registers(hackrf_device* device)
+{
+	uint16_t register_number;
+	int result = HACKRF_SUCCESS;
+
+	for (register_number = 1; register_number <= 5; register_number++) {
```
mndza: I left out the bulk read function for FPGA registers intentionally: we don't know the number of registers any gateware might have. I'm still not sure about this.

firmware/common/fpga.c
```diff
@@ -22,226 +22,76 @@
...
-	bool q_in_range = (zcs_q > expected_zcs * 0.95) && (zcs_q < expected_zcs * 1.05);
-
-	// Update selftest result.
-	selftest.xcvr_loopback_ok = i_in_range && q_in_range;
-	if (!selftest.xcvr_loopback_ok) {
-		selftest.report.pass = false;
+void fpga_nco_enable(fpga_driver_t* const drv, const bool enable)
+{
+	if (enable) {
+		set_FPGA_STANDARD_CTRL_DC_BLOCK(drv, 1);  // RX DC block
```
mndza: Enabling the TX NCO does not involve enabling RX DC block

host/hackrf-tools/src/hackrf_debug.c
mndza: We could find a way to do it, yeah

firmware/common/fpga.c
mndza: Well, I wouldn't join both features in a single one. In the selftest, we enable DC block and then enable TX NCO.

firmware/common/fpga.c
```diff
@@ -22,226 +22,84 @@
...
-
-struct fpga_image_read_ctx {
-	uint32_t addr;
-	size_t next_block_sz;
-	uint8_t init_flag;
-	uint8_t buffer[4096 + 2];
-};
-
-static size_t fpga_image_read_block_cb(void* _ctx, uint8_t* out_buffer)
+void fpga_setup(fpga_driver_t* const drv)
```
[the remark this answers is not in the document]
mndza: I tried to make this distinction in `hackrf_debug`:
```
	-g, --gateware: target gateware registers
	-P, --fpga <n>: load the n-th bitstream to the FPGA
```

PR #1639, opened 2026-01-07
gateware: generic SGPIOInterface, simpler capture mgmt, fix resampler bugs

### LINKED ISSUE #1629: Praline: spectrum inverts randomly when changing sample rate

Spectrum inverts randomly when changing sample rate since 99f96d24981862a46ce90905efc5eadf5ec464ec

TX or RX with GNU Radio and gr-osmosdr. Adjust sample rate with slider.

_No response_

mndza:
Also fixes #1629


mossmann:
This does not fix #1629 for me. I can get the spectrum to invert by repeatedly adjusting the sample rate. It sometimes gets into a state where crossing a decimation threshold (usually at 2.5 Msps) reliably flips the spectrum for a while.

This may have been random, but it seemed like it was easier to reproduce the problem after changing the starting sample rate from 2.5 Msps to 20 Msps in the flowgraph.

Issue #906, opened 2021-07-05
store settings in firmware


mossmann:
Many of the functions in libhackrf affect the state of the hardware without those settings being stored in RAM by the firmware. For example, `hackrf_set_lna_gain()` passes a numeric setting to the HackRF which the firmware converts to a setting for the MAX2837. The firmware then stores the value only in a register of the MAX2837.

Retrieving such values from the registers of various components is non-trivial and has not been implemented. In order to facilitate the creation of "get" functions to complement "set" functions (#129) and potentially for other purposes, it would be helpful if the firmware stored such configuration settings in RAM when those settings take effect. Providing the ability for the host to retrieve the settings would enable user software to report the state of the hardware (#578).

In some cases, user-requested settings may not be precisely honored by the firmware, but we currently provide the user no way to confirm this (#524).

I propose a new static settings structure in `firmware/common/settings.c` that stores settings configured by all of the following libhackrf functions:
```
hackrf_set_amp_enable
hackrf_set_antenna_enable
hackrf_set_baseband_filter_bandwidth
hackrf_set_clkout_enable
hackrf_set_freq
hackrf_set_freq_explicit
hackrf_set_hw_sync_mode
hackrf_set_lna_gain
hackrf_set_operacake_ports
hackrf_set_operacake_ranges
hackrf_set_sample_rate
hackrf_set_sample_rate_manual
hackrf_set_txvga_gain
hackrf_set_ui_enable
hackrf_set_vga_gain
``

Additionally we should move transceiver state and sweep mode state into this structure, and we should follow this pattern for future firmware features that are configured over USB.

Potentially this structure could be used in the future to provide a way to load and store all settings at once.

PR #1648, opened 2026-01-20
Radio configuration register interface

### LINKED ISSUE #1681: Praline tuning error on sample rate change

With HackRF Pro (Praline), changing the sample rate without changing frequency during RX can result in a tuning offset. This was introduced in a5af0edabbbb82921186d54828c0b48c51036d72.

Workaround: Readjust frequency after changing sample rate.

Start RX. Adjust sample rate.

_No response_

### LINKED ISSUE #1683: Mixer PLL fails to lock

intermittent - occurring irregularly

The RFFC5072 PLL can fail to lock for multiple reasons, resulting in incorrect and drifting frequency tuning. This affects all platforms except rad1o. These problems were greatly exacerbated by various changes in release 2026.01.1.

A large majority of cases were fixed in #1658 (release 2026.01.3), but occasional lock failures may still occur due to a concurrency problem in firmware.

Start RX. Observe unexpected region of spectrum. Observe frequency drift (which may not be very apparent if gains and signal level are low).

_No response_

mossmann:
This implements a new firmware API for unified control of the radio section with a register interface. Requested configuration changes are written to the active register bank. Settings in the active bank are periodically applied and written to the applied bank. Additional banks for specific operating modes store settings that will be copied to the active bank and then applied when switching to that operating mode. For example, settings in the RX bank are copied and applied when starting RX. This effectively implements most of #906.

This initial implementation replaces the underlying structure without changing function from the perspective of host software. It aims to enable future work:

- Make all radio settings retrievable by the host. (Some of this is already done, but additional work will enable, for example, retrieval of the precise tuned frequency instead of the requested frequency.)
- Consolidate radio elements under radio.c, moving them out of hackrf_core.c.
- Update libhackrf and documentation to make the register interface useful to host software.
- Implement additional radio configuration modes that use alternative FPGA bitstreams on HackRF Pro.

Additionally, by moving the application of radio configuration changes outside of interrupt handlers, this eliminates a class of concurrency bug that can cause tuning failure or other configuration errors.

Fix #1681
Fix #1683


martinling (changes requested):

martinling:
I was just doing some testing with the rad1o badge, and found that sweep mode has been broken there since commit b1bfc246bf8a7459ceec42089b48a7e63a439474. I don't know the cause yet.

PR #1657, opened 2026-01-27
Unified Firmware

antoinevg:
This PR introduces a new `BOARD` target called `UNIVERSAL` that can be used to compile a combined firmware that supports both HackRF One and HackRF Pro hardware.

## Conventions

It introduces a new pattern to the code base for working with board support.

Previously, we used a mix of patterns of the form `#ifdef HACKRF_ONE` or `#ifndef PRALINE` in conjunction with `platform_detect.detected_platform()` to manage board differences.

This PR applies the following new patterns across the codebase:

* Conditional directives now only serve as guards for what code gets linked into the firmware for a given board.
* To control code execution always use `if (detected_platform() == BOARD_ID_x)` or `switch (detected_platform())`
* Conditional directives of the form `#ifdef SOME_BOARD` become `#if defined(SOME_BOARD)` or `#ifndef SOME_BOARD` becomes `#if !defined(SOME_BOARD)`. This keeps usage consistent with the next point.
* Specifically, for `HACKRF_ONE` and `PRALINE` always include the `UNIVERSAL` target to ensure the code will be present in the combined firmware binary. For example: `#if defined(PRALINE) || defined(UNIVERSAL)`
* Be careful of `#else`, this caught me a few times: 
> ```
> #if defined(PRALINE) || defined(UNIVERSAL)`
>     ... some code ...
> #else
>     ... this code now gets left out of UNIVERSAL ...
> #endif
> ```
> Make sure to use this form instead:
> ```
> #if defined(PRALINE) || defined(UNIVERSAL)`
>     ... some code ...
> #endif
> #if !defined(PRALINE) || defined(UNIVERSAL)`
>     ... this code is now also included in UNIVERSAL ...
> #endif
> ```

## Compilation

```
mkdir firmware/build
cd firmware/build
cmake .. -DBOARD=UNIVERSAL
make
```

## Size impact

Single-board firmware are slightly bigger as a result:

```
PRALINE:     was: 240288  => now:  243336  difference: 3048
HACKRF_ONE:  was:  45920  => now:   48672  difference: 2752
RAD1O:       was:  75244  => now:   77548  difference: 2304
JAWBREAKER:  was:  38060  => now:   39988  difference: 1928
JELLYBEAN:   was:    n/a  => now:   39204
```

The size increase is mostly due to: 

* `platform_scu:   968 bytes`
* `platform_gpio: 1044 bytes`

…with the remaining being due to runtime platform detection across the rest of the code-base.

The combined firmware comes to `258376` bytes.

## Testing

This PR has been tested with both single-board and combined firmware using `hackrf.git/ci-scripts/hackrf-test.py` and `praline-factory-test.git/bin/hackrf_test.py` in the following configurations:

#### `EUT`: Hackrf One r4, `TESTER`: HackRF One r10c Portapack H4M clone from SDRStore

```
python hackrf_test.py --unattended --ci --eut $EUT --tester $TESTER --rev any -f $FIRMWARE -L /tmp/log
```

#### `EUT`: Hackrf Pro, TESTER: `HackRF One r4`

```
python hackrf_test.py --factory -p --eut $EUT --tester $TESTER --rev any -f $FIRMWARE -L /tmp/log --tcxo
```

## Future Work

Now that conditional compilation directives are no longer used to control code execution we have the opportunity to consider whether we should follow the example of `firmware/common/rad1o/` and add `firmware/common/hackrf_one/` and `firmware/common/praline/` directories.

---

- [x] Finish up proof of concept for `rf_path.c`
- [x] Add `max2831` support to `max283x`
- [x] Refactor remaining gpio's into `platform_gpio.c`
- [x] Refactor remaining scu's into `platform_scu.c`
- [x] Figure out how to handle config & drivers that need to setup early
- [x] Figure out how to handle LED's for `platform_detect.c`
- [x] Refactor remaining `#ifdef`s
- [x] Get `PRALINE` binary booting on `HACKRF_ONE`
- [x] Test. Test. Test.
- [x] Clean up `#include`s
- [x] Fix MCU LED not lighting up on HackRF Pro
- [x] Fix supported platforms text in `hackrf_info`
- [x] Add  platform compilation conditionals back

### DIFF
--- a/firmware/common/max283x.c
+++ b/firmware/common/max283x.c
@@ -385,6 +385,34 @@ void max283x_rx(max283x_driver_t* const drv)
 	}
 }
 
+/* Set MAX2831 receiver high-pass filter corner frequency in Hz */
+void max283x_set_rx_hpf_frequency(
+	max283x_driver_t* const drv,
+	const max283x_rx_hpf_freq_t freq)
+{
+#if !defined(PRALINE)
+	(void) freq;
+#endif
+
+	switch (drv->type) {
+#if defined(PRALINE) || defined(UNIVERSAL)
+	case MAX2831_VARIANT:
+		max2831_set_rx_hpf_frequency(
+			&drv->drv.max2831,
+			(max2831_rx_hpf_freq_t) freq);
+		break;
+#endif
+#if !defined(PRALINE) || defined(UNIVERSAL)
+	case MAX2837_VARIANT:
+		// unsupported
+		break;
+	case MAX2839_VARIANT:
+		// unsupported
+		break;
+#endif
+	}
+}
+
 void max283x_tx_calibration(max283x_driver_t* const drv)
 {
 	switch (drv->type) {
@@ -395,10 +423,10 @@ void max283x_tx_calibration(max283x_driver_t* const drv)
 #endif
 #if !defined(PRALINE) || defined(UNIVERSAL)
 	case MAX2837_VARIANT:
-		// unsupported - uses max283x_set_mode
+		// unsupported - use max283x_set_mode instead
 		break;
 	case MAX2839_VARIANT:
-		// unsupported - uses max283x_set_mode
+		// unsupported - use max283x_set_mode instead
 		break;
 #endif
 	}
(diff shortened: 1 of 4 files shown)


mossmann:
Regarding the `#else` footgun: Did you consider defining both `PRALINE` and `HACKRF_ONE` when `UNIVERSAL` is defined (and then testing only for `PRALINE`, not both `PRALINE` and `UNIVERSAL` in typical usage)?

antoinevg:
> Regarding the `#else` footgun: Did you consider defining both `PRALINE` and `HACKRF_ONE` when `UNIVERSAL` is defined (and then testing only for `PRALINE`, not both `PRALINE` and `UNIVERSAL` in typical usage)?

It does make some cases simpler but still leaves cases where `UNIVERSAL` is needed so I went with consistency and keeping things explicit instead.

It was not fun adding back all the conditional directives after getting rid of them _but_ we are now in a good position for a relatively straight-forward future refactor where we can lose most of the conditional directives in the source code in favor of splitting board-specific code into their own compilation units.

mossmann:
replaced by #1715

PR #1658, opened 2026-01-27
Fix mixer lock

mossmann:
The RFFC5072 PLL can fail to lock for multiple reasons, resulting in incorrect and drifting frequency tuning.

One cause is enabling the part with frequency registers set to zero, something we've always done. Fixing this reduced the incidence of PLL lock failure on cold start. Another thing that reduced incidence of PLL lock failure on cold start is enabling reference oscillator standby. With both fixes together, I was no longer able to reproduce PLL lock failure on cold start.

Another cause is that the default registers recommended in the RFFC5072 documentation configure too short/aggressive warm-up and calibration parameters. Updated registers produced by the vendor's configuration software extend the tuning time and improve tuning reliability.

### DIFF
--- a/firmware/common/rffc5071.c
+++ b/firmware/common/rffc5071.c
@@ -40,17 +40,17 @@
 #include <libopencm3/lpc43xx/scu.h>
 #include "hackrf_core.h"
 
-/* Default register values. */
+/* Default register values from vendor documentation or software. */
 static const uint16_t rffc5071_regs_default[RFFC5071_NUM_REGS] = {
-	0xbefa, /* 00 */
+	0xfffb, /* 00 */
 	0x4064, /* 01 */
 	0x9055, /* 02 */
 	0x2d02, /* 03 */
-	0xacbf, /* 04 */
-	0xacbf, /* 05 */
+	0xb0bf, /* 04 */
+	0xb0bf, /* 05 */
 	0x0028, /* 06 */
 	0x0028, /* 07 */
-	0xff00, /* 08 */
+	0xfc06, /* 08 */
 	0x8220, /* 09 */
 	0x0202, /* 0A */
 	0x0400, /* 0B */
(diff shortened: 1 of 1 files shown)


miek:
If possible, just to make it easier to follow exactly what changed and limit the amount of magic in the default registers update, I'd like the commit history changed a bit:
 * Modify the TVCO commit to make the change in `rffc5071_setup` directly (rather that re-doing it in the default registers update commit)
 * Similarly, move the dithering disable to `rffc5071_setup` in its own commit.
 * As far as I can tell, there wasn't actually an inadvertent change to the DSM modulator (#1567 changed register 0x12 from 0x2a20 to 0x3a20, and then it was 0x2a20 a long way back in the history), so remove the DSM modulator commit.

(I've made those changes here if you want to pull that: https://github.com/miek/hackrf/commits/fix-mixer-lock/)

mossmann:
>     * As far as I can tell, there wasn't actually an inadvertent change to the DSM modulator ([Tuning improvements #1567](https://github.com/greatscottgadgets/hackrf/pull/1567) changed register 0x12 from 0x2a20 to 0x3a20, and then it was 0x2a20 a long way back in the history), so remove the DSM modulator commit.

#1567 changed SDM from `0b10` to `0b11`. This changed the modulator from "3rd order MASH 1-1-1" (in which DSM_MODE was ignored) to "DSM modulator as defined by mode, dith, fm and dmode fields" (in which DSM_MODE is honored, changing from 3rd order to 5th order).

miek:
> > ```
> > * As far as I can tell, there wasn't actually an inadvertent change to the DSM modulator ([Tuning improvements #1567](https://github.com/greatscottgadgets/hackrf/pull/1567) changed register 0x12 from 0x2a20 to 0x3a20, and then it was 0x2a20 a long way back in the history), so remove the DSM modulator commit.
> > ```
> 
> #1567 changed SDM from `0b10` to `0b11`. This changed the modulator from "3rd order MASH 1-1-1" (in which DSM_MODE was ignored) to "DSM modulator as defined by mode, dith, fm and dmode fields" (in which DSM_MODE is honored, changing from 3rd order to 5th order).

Ahh OK, yeah, I follow that now. I've put that change back in on my branch.

mossmann:
@miek I've copied your branch. Ready for final review!

PR #1666, opened 2026-02-03
Toolchain 9.2.1 support fix

stafur:
Added -B arm flag to firmware/hackrf_usb/CMakeLists.txt to support builds with v9.2.1 toolchains.


mossmann:
Thank you!

PR #1516, opened 2024-12-13
fix compilation warnings

gullradriel:
That PR is fixing all the following warnings:
```
In file included from ~/portapack-mayhem/hackrf/firmware/common/usb.c:26:
~/portapack-mayhem/hackrf/firmware/common/usb.h:32:1: warning: function declaration isn't a prototype [-Wstrict-prototypes]
   32 | void usb_peripheral_reset();
      | ^~~~
~/portapack-mayhem/hackrf/firmware/common/usb.h:33:1: warning: function declaration isn't a prototype [-Wstrict-prototypes]
   33 | void usb_phy_enable();
      | ^~~~
In file included from ~/portapack-mayhem/hackrf/firmware/common/usb.c:32:
~/portapack-mayhem/hackrf/firmware/libopencm3/include/libopencm3/lpc43xx/m4/nvic.h:45: warning: "NVIC_SGPIO_IRQ" redefined
   45 | #define NVIC_SGPIO_IRQ 31
      | 
In file included from ~/portapack-mayhem/hackrf/firmware/libopencm3/include/libopencm3/dispatch/nvic.h:30,
                 from ~/portapack-mayhem/hackrf/firmware/libopencm3/include/libopencm3/cm3/nvic.h:135,
                 from ~/portapack-mayhem/hackrf/firmware/libopencm3/include/libopencm3/lpc43xx/m4/nvic.h:9,
                 from ~/portapack-mayhem/hackrf/firmware/common/usb.c:32:
~/portapack-mayhem/hackrf/firmware/libopencm3/include/libopencm3/lpc43xx/m0/nvic.h:34: note: this is the location of the previous definition
   34 | #define NVIC_SGPIO_IRQ 19
      | 
In file included from ~/portapack-mayhem/hackrf/firmware/common/usb.c:32:
~/portapack-mayhem/hackrf/firmware/libopencm3/include/libopencm3/lpc43xx/m4/nvic.h:50: warning: "NVIC_PIN_INT4_IRQ" redefined
   50 | #define NVIC_PIN_INT4_IRQ 36
      | 
In file included from ~/portapack-mayhem/hackrf/firmware/libopencm3/include/libopencm3/dispatch/nvic.h:30,
                 from ~/portapack-mayhem/hackrf/firmware/libopencm3/include/libopencm3/cm3/nvic.h:135,
                 from ~/portapack-mayhem/hackrf/firmware/libopencm3/include/libopencm3/lpc43xx/m4/nvic.h:9,
[...]

PR #1667, opened 2026-02-06
Firmware: Enable `-Wstrict-prototypes` and fix remaining warnings

martinling:
Declaring a function `f()` with no arguments is not the same as declaring it with `f(void)`, for reasons that have been deprecated since C89.

Some instances of this were fixed in #1516. This finishes the job, and enables the relevant warning so that we don't introduce more.


mossmann:
Is there any argument for or against making this change also in function definitions (e.g. operacake_sctimer_init() in operacake_sctimer.c)?

martinling:
I intended to make it everywhere, but I missed `operacake_sctimer_init()` and some others because the warning didn't trigger for them, due to them having a valid prototype in the header declaration.

I've now caught those stragglers by grepping the firmware source for `()$`, which works reliably thanks to #1112.

mossmann (approved):
Thank you!

miek:
> I intended to make it everywhere, but I missed `operacake_sctimer_init()` and some others because the warning didn't trigger for them, due to them having a valid prototype in the header declaration.

fwiw, I think `-Wold-style-definition` would catch examples like that

PR #1668, opened 2026-02-06
Rename `struct gpio_t` to `struct gpio` to distinguish from `gpio_t`

martinling:
We've been using `gpio_t` as both the name of a `struct`, and a typedef for a pointer to that `struct`.

In C that's possible because `struct gpio_t` names a different type from `gpio_t` , but it's confusing, and causes problems when trying to include the types elsewhere.

### DIFF
--- a/firmware/common/gpio.h
+++ b/firmware/common/gpio.h
@@ -25,7 +25,7 @@
 
 #include <stdbool.h>
 
-typedef const struct gpio_t* gpio_t;
+typedef const struct gpio* gpio_t;
 
 void gpio_init(void);
 void gpio_set(gpio_t gpio);
--- a/firmware/common/gpio_lpc.h
+++ b/firmware/common/gpio_lpc.h
@@ -32,7 +32,7 @@
  * register #defines and API declarations into separate header files.
  */
 
-typedef struct gpio_port_t {
+typedef struct gpio_port {
 	volatile uint32_t dir; /* +0x000 */
 	uint32_t _reserved0[31];
 	volatile uint32_t mask; /* +0x080 */
@@ -48,7 +48,7 @@ typedef struct gpio_port_t {
 	volatile uint32_t inv; /* +0x300 */
 } gpio_port_t;
 
-struct gpio_t {
+struct gpio {
 	const uint32_t mask;
 	gpio_port_t* const port;
 	volatile uint32_t* const gpio_w;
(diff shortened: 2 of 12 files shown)


martinling:
@antoinevg This will conflict badly with your work in #1657, so if you'd rather include this change in that PR somehow, I can drop this one.

antoinevg:
> @antoinevg This will conflict badly with your work in #1657, so if you'd rather include this change in that PR somehow, I can drop this one.

Thank you @martinling, good call!

Commit with the change is here for perusal and/or any feedback: fd6ab3bbbf9a2fdb0b4594a24474d5b0de5a0a9e

martinling:
@antoinevg You have the same problem in that commit, just with both types being named `gpio` instead of both named `gpio_t`. This causes problems when the code is used elsewhere, like:

```
firmware/common/gpio.h:28:30: error: conflicting declaration 'typedef const struct gpio_t* gpio_t'
   28 | typedef const struct gpio_t* gpio_t;
      |                              ^~~~~~
firmware/common/gpio.h:28:22: note: previous declaration as 'struct gpio_t'
   28 | typedef const struct gpio_t* gpio_t;
      |                      ^~~~~~

```

antoinevg:
Oh ***doh***, it's worse than I thought… Tx!

Updated commit: 10710f09867ff0826febf2a5d590553196433ab7

martinling:
@antoinevg Yup that looks good. I'll close this one.

antoinevg:
Re-opened and rebased against `main`.

antoinevg:
We originally had two types sharing the name `gpio_t`:

1. A struct called `gpio_t`

```
struct gpio {
    const uint32_t mask;
    gpio_port_t* const port;
    volatile uint32_t* const gpio_w;
};
```

2. …and a `typedef` which represented a pointer to a `const struct gpio_t`:

```
typedef const struct gpio_t* gpio_t;
```

The struct has been renamed to `gpio` so now `gpio_t` _only_ means a `const` pointer to a `struct gpio`:

```
struct gpio {
    const uint32_t mask;
    gpio_port_t* const port;
    volatile uint32_t* const gpio_w;
};

typedef const struct gpio* gpio_t;
```

The `gpio_t`'s in `portapack.c` are meant to be `const` pointers to a `struct gpio` :-)

(See exchange between me and @martinling  above for my shared initial confuzzlement)

PR #1676, opened 2026-02-16
gateware: rearrange registers, DAC clock domain, little-endian extended samples

mndza:


mossmann:
Discovered #1681 while testing this PR, but it turned out to be unrelated.

mossmann (changes requested):
I'm seeing a TX frequency inversion introduced when changing to a high sample rate (>18 Msps).

firmware/common/fpga_regs.def
```diff
@@ -53,22 +53,23 @@ const uint8_t n = r;
...
 __MREG__(FPGA_STANDARD_CTRL_TRIGGER_EN, 1, 7, 1)
 
 /* REG 02 (2): RX_DECIM */
 __MREG__(FPGA_STANDARD_RX_DECIM, 2, 0, 3)
 
-/* REG 03 (3): TX_CTRL */
-__MREG__(FPGA_STANDARD_TX_CTRL, 3, 0, 1)
-__MREG__(FPGA_STANDARD_TX_CTRL_NCO_EN, 3, 0, 1)
+/* REG 03 (3): RX_PSTEP */
+__MREG__(FPGA_STANDARD_RX_PSTEP, 3, 0, 8)
```
antoinevg: Currently we're only using 2 bits of the register, are we going to end up using the extra 6 bits or can we make it 2 bits long in the gateware and here?

mossmann:
>     * Praline r1.0.0: `FAIL 12: Loopback test failed` (loopback also fails on selftest)

I'll investigate this further, but I think it is due to a hardware design change between r1.0.0 and r1.2.1.  Can you please also test on r1.2.1?

[edit: Nevermind! I see you did test on r1.2.1, identified as "HackRF Pro".]

mossmann:
@antoinevg I've addressed Praline r1.0 compatibility.

antoinevg:
> @antoinevg I've addressed Praline r1.0 compatibility.

👍 selftest good here now.

Factory test is giving a `FAIL 823: AMP rxhp1 clipping` but that's always been the case since I started running it on `r1.0.0` a month or so back.

PR #1688, opened 2026-02-23
Improve sample rate precision

mossmann:
- Fix baseband bandwidth and sample rate configuration bugs which break Praline HITL CI
- Configure sample rate more precisely with fixed-point type
- Use external callback to set radio sample rate


mossmann:
Set to draft again because I noticed that the UI is broken and TX spectrum looks wrong.

PR #1689, opened 2026-02-25
Add radio configuration update callback

mossmann:
Depends on #1708 

A new callback replaces and centralizes UI and Opera Cake update calls scattered throughout the code.

### DIFF
--- a/firmware/common/rf_path.c
+++ b/firmware/common/rf_path.c
@@ -32,7 +32,6 @@
 #include "max5864.h"
 #include "mixer.h"
 #include "platform_detect.h"
-#include "sgpio.h"
 #if defined(HACKRF_ONE) || defined(RAD1O) || defined(PRALINE)
 	#include "platform_scu.h"
 #endif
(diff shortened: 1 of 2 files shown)


mossmann:
I believe that the blinky build is failing due to an include dependency problem. I'll see if #1657 solves it.

mossmann:
The blinky build was actually failing because I shouldn't have put the new function into hackrf_core.c. I moved it to hackrf_usb.c.

martinling:
I figured out what was going wrong and have pushed fixes to this branch.

mossmann:
@martinling Thank you! I believe I had fixed both those problems previously but somehow lost my fixes.

martinling:
I'd thought with that logic fixed that #1697 might work, but that's now differently broken.

mossmann:
I incorporated @martinling's changes, fixed some additional bugs, and rebased. I set this back to draft because I found an additional bug that is not yet fixed:

The PortaPack UI does not correctly display the image reject filter state. This is because the radio configuration registers do not yet store the applied image reject filter state. I have some work in progress that stores this and other tuning registers, but it needs some work to bring up to date with main. I think it would be best to finish that work and merge it before this.

mossmann:
The required storage of image reject filter state was implemented in #1708.

PR #1702, opened 2026-03-17
Change Praline CLKIN default to P1_CLKIN

### LINKED ISSUE #1624: Add Praline P1/P2 support to hackrf_clock
### What feature would you like to see and why?

P1 and P2 multiplexers can be controlled with hackrf_debug, but we would also like more user-friendly options in hackrf_clock. This should support basic CLKIN, CLKOUT, and trigger IO functions.

mossmann:
Enable P1_CLKIN by default and disconnect P22_CLKIN by default.

Close #1624 

[edit: Originally this changed the P1 default from P1_CLKIN to P22_CLKIN, matching the default behavior of HackRF One. See the following discussion.]

### DIFF
--- a/ci-scripts/hackrf_pro_test.py
+++ b/ci-scripts/hackrf_pro_test.py
@@ -663,9 +663,6 @@ class HackRF:
             self.rf_test_cases = praline_r110_rf_test_cases
 
     def clkin(self):
-        if self.name == "EUT" and self.id == "5":
-            debug = subprocess.run([self.bin_dir + "/hackrf_debug", "-d", self.serial,
-                                    "-C", "0"])
         command = subprocess.run([self.bin_dir + "/hackrf_clock", "-i", "-d",
             self.serial], capture_output=True, encoding="utf-8",
             timeout=TIMEOUT)
--- a/docs/source/external_clock_interface.rst
+++ b/docs/source/external_clock_interface.rst
@@ -9,12 +9,18 @@ HackRF Pro
 
 HackRF Pro has two configurable SMA ports, P1 and P2. By default, P1 is configured as CLKIN and P2 as CLKOUT. The default behaviour of these signals is as described for HackRF One below.
 
+A second CLKIN signal is available on header P22 pin 2. Unlike HackRF One, HackRF Pro's P22_CLKIN is a separate signal from P1_CLKIN. To enable P22_CLKIN instead of P1_CLKIN use ``hackrf_clock -c p22``.
+
+Various internal signals can be connected to P1 or P2 instead of the default CLKIN and CLKOUT signals. Use ``hackrf_clock -1`` or ``hackrf_clock -2`` to select a different signal.
+
 HackRF One
 ~~~~~~~~~~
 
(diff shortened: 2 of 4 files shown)


martinling (commented):
This will have the side effect that if a clock is connected to P22, it will also appear as an output on P1. This was true on HackRF One but arguably not a great default.

I think this PR should also update the block diagram, which currently shows the default path as being from P1 through to the Si5351 only.

mossmann:
I agree that this should update the block diagram, but the CLKIN default in the block diagram was already incorrect. (It shows P1_CLKIN as the default while the default in firmware is P22_CLKIN.)

This PR makes the default behavior consistent with our external clock input documentation.

martinling (approved):
Yes, the block diagram was wrong before. I'm not sure if I was referencing the unpowered state of the CLKIN_CTRL switch, or if I just assumed that the default usage of P1 being CLKIN meant that CLKIN_CTRL would be set accordingly.

mossmann:
After further discussion, we would like to take this in a different direction, making P1_CLKIN work by default and disconnecting P22_CLKIN by default.

In every case that we know of, external clocks connected to P22 are no better than the internal TCXO on HackRF Pro, so it is reasonable to ignore them by default. When someone connects an external clock to P1, it is much more likely to be a superior clock or to be part of a multi-device setup where we do not want to ignore the external clock by default. Connecting P22_CLKIN to P1_CLKIN by default would make it easy for a P1_CLKIN user to connect two clock outputs together, potentially damaging one or both clock sources.

This will, however, mean that the default behavior on HackRF Pro will differ from HackRF One which has no ability to isolate P22_CLKIN from P1_CLKIN. This change will require an update to `hackrf_clock` (https://github.com/greatscottgadgets/hackrf/issues/1624) and to documentation.

antoinevg (approved):
Code looks good, tested as follows:

```bash
> export HACKRF_ONE=000000000000000057b068dc22726e63
> export PRALINE=0000000000000000977c64de31426d13

# -- HACKRF_ONE P2 connected to PRALINE P1

# reset firmware
> hackrf_spiflash -d $HACKRF_ONE -R
> hackrf_spiflash -d $PRALINE -R

> hackrf_clock -d $PRALINE -i
CLKIN status: no clock signal detected

> hackrf_clock -d $HACKRF_ONE -o 1 # enable HACKRF_ONE clock out
> hackrf_clock -d $PRALINE -i
CLKIN status: clock signal detected

# -- HACKRF_ONE P2 connected to PRALINE P22_2

# reset firmware
> hackrf_spiflash -d $HACKRF_ONE -R
> hackrf_spiflash -d $PRALINE -R

> hackrf_clock -d $PRALINE -i
CLKIN status: no clock signal detected

> hackrf_clock -d $HACKRF_ONE -o 1 # enable HACKRF_ONE clock out
> hackrf_clock -d $PRALINE -i
CLKIN status: no clock signal detected

> hackrf_clock -d $PRALINE -c p22 # select P22 CLKIN
> hackrf_clock -d $PRALINE -i
CLKIN status: clock signal detected

```

martinling (approved):

PR #1703, opened 2026-03-20
Replace include guards with `#pragma once`

antoinevg:
Depends on #1701

---

This PR removes include guards in the style of:

```c
#ifndef __SOME_FILE_H__
#define __SOME_FILE_H__

...

#endif // __SOME_FILE_H__
```

…and replaces them with:

```c

#pragma once 

...
```

### DIFF
--- a/firmware/common/adc.h
+++ b/firmware/common/adc.h
@@ -19,12 +19,9 @@
  * Boston, MA 02110-1301, USA.
  */
 
-#ifndef __ADC_H__
-#define __ADC_H__
+#pragma once
 
 #include <stdint.h>
 
 uint16_t adc_read(uint8_t pin);
 void adc_off(void);
-
-#endif // __ADC_H__
--- a/firmware/common/bitband.h
+++ b/firmware/common/bitband.h
@@ -20,8 +20,7 @@
  * Boston, MA 02110-1301, USA.
  */
 
-#ifndef __BITBAND_H__
-#define __BITBAND_H__
+#pragma once
 
 #include <stdint.h>
 
@@ -37,5 +36,3 @@ void peripheral_bitband_clear(
 uint32_t peripheral_bitband_get(
 	volatile void* const peripheral_address,
 	const uint_fast8_t bit_number);
-
-#endif //__BITBAND_H__
--- a/firmware/common/clkin.h
+++ b/firmware/common/clkin.h
@@ -19,12 +19,9 @@
  * Boston, MA 02110-1301, USA.
  */
 
-#ifndef __CLKIN_H__
-#define __CLKIN_H__
+#pragma once
 
 #include <stdint.h>
 
 void clkin_detect_init(void);
 uint32_t clkin_frequency(void);
-
-#endif //__CLKIN_H__
--- a/firmware/common/cpld_jtag.h
+++ b/firmware/common/cpld_jtag.h
@@ -19,8 +19,7 @@
  * Boston, MA 02110-1301, USA.
  */
 
-#ifndef __CPLD_JTAG_H__
-#define __CPLD_JTAG_H__
+#pragma once
 
 #include <stdint.h>
 
@@ -59,5 +58,3 @@ int cpld_jtag_program(
 	unsigned char* const buffer,
 	refill_buffer_cb refill);
 unsigned char cpld_jtag_get_next_byte(void);
-
-#endif //__CPLD_JTAG_H__
--- a/firmware/common/cpld_xc2c.h
+++ b/firmware/common/cpld_xc2c.h
@@ -20,8 +20,7 @@
  * Boston, MA 02110-1301, USA.
  */
 
-#ifndef __CPLD_XC2C_H__
-#define __CPLD_XC2C_H__
+#pragma once
 
 #include <stdbool.h>
 #include <stdint.h>
@@ -69,5 +68,3 @@ bool cpld_xc2c64a_jtag_sram_verify(
 extern const cpld_xc2c64a_program_t cpld_hackrf_program_sram;
 extern const cpld_xc2c64a_verify_t cpld_hackrf_verify;
 extern const cpld_xc2c64a_row_addresses_t cpld_hackrf_row_addresses;
-
-#endif /*__CPLD_XC2C_H__*/
--- a/firmware/common/crc.h
+++ b/firmware/common/crc.h
@@ -20,8 +20,7 @@
  * Boston, MA 02110-1301, USA.
  */
 
-#ifndef __CRC_H__
-#define __CRC_H__
+#pragma once
 
 #include <stddef.h>
 #include <stdint.h>
@@ -35,5 +34,3 @@ typedef struct {
 void crc32_init(crc32_t* const crc);
 void crc32_update(crc32_t* const crc, const uint8_t* const data, const size_t byte_count);
 uint32_t crc32_digest(const crc32_t* const crc);
-
-#endif //__CRC_H__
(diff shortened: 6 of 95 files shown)


gullradriel:
#pragma once tells the compiler "only include this file once", but the compiler has to figure out what "this file" means. It typically does that using the file's path or inode. 
If the same header is reachable via two paths (a symlink, a bind mount, a copied include directory), the compiler may see them as two different files and include it twice, defeating the whole point. 
Header guards don't have this problem because they work purely in the preprocessor's token space, the guard macro is either defined or it isn't, regardless of how you got to the file.
I would call it an unnecessary AI code bad move.

martinling:
@gullradriel

> If the same header is reachable via two paths (a symlink, a bind mount, a copied include directory), the compiler may see them as two different files and include it twice, defeating the whole point.

If you have two headers with the same name, in directories that are both in the compiler's list of include paths, like
`inc/a/api.h` and `inc/b/api.h`, with `-Iinc/a -Iinc/b` passed to the compiler; then `#include <api.h>` is only ever going to include `a/api.h`, because that's the `api.h` found first in the search path. It won't include both, regardless of whether `#pragma once` or traditional header guards are used.

To include _both_ `api.h` headers, you'd have to make it possible to distinguish them by the paths used in the `#include` statements. E.g. pass `-Iinc` to the compiler, and then write:

```c
#include <a/api.h>
#include <b/api.h>
```

In which case, it's pretty clear that you're actually _intentionally_ including both headers.

> Header guards don't have this problem because they work purely in the preprocessor's token space

That's actually exactly the problem with them. Because there's no guaranteed way to uniquely name them, it's quite likely that if `a` and `b` are unrelated projects, both `a/api.h` and `b/api.h` might choose _the same_ guard name, such as `API_H`. The more generic the name of the header, the more likely this is to happen.

In fact, this is exactly the scenario that led me to ask @antoinevg for this change. I was working on our Portapack firmware project that @mossmann mentioned in https://github.com/portapack-mayhem/mayhem-firmware/issues/2957#issuecomment-3874193138, and I ended up needing to include both `gpio.h` from the HackRF source, and `gpio.hpp` from the PortaPack source. Both of them used the same guard name, `__GPIO_H__`, making it impossible to use both at once. At that point the only option was to modify one of the projects to prevent the clash.

This is where `#pragma once` is a better solution, precisely because it _doesn't_ depend on choosing unique names. It also avoids a lot of other mistakes that are common with traditional guards, like:
- clashes with system headers (officially, all identifiers starting with `__` are reserved to the C implementation!)
- inconsistency about how to name guards within the project (e.g. we have a mix of `__NAME_H` and `__NAME_H__`)
- forgetting to change a guard name when renaming a file, or copying code from one file to another
- mismatching the `#ifndef` and the `#define`, (as I [fixed recently in our libopencm3 submodule](https://github.com/mossmann/libopencm3/pull/33)).

The main issue with `#pragma once` was that it wasn't always supported by all compilers, but that hasn't been the case for [a very long time now](https://en.wikipedia.org/wiki/Pragma_once#Portability) - it's been in GCC since 3.4 in 2004, in clang from the start, and in everything else for ages too by now.

mossmann (approved):
Excellent! Thank you!

gullradriel:
> @gullradriel
> 
> > If the same header is reachable via two paths (a symlink, a bind mount, a copied include directory), the compiler may see them as two different files and include it twice, defeating the whole point.
> 
> If you have two headers with the same name, in directories that are both in the compiler's list of include paths, like `inc/a/api.h` and `inc/b/api.h`, with `-Iinc/a -Iinc/b` passed to the compiler; then `#include <api.h>` is only ever going to include `a/api.h`, because that's the `api.h` found first in the search path. It won't include both, regardless of whether `#pragma once` or traditional header guards are used.
> 
> To include _both_ `api.h` headers, you'd have to make it possible to distinguish them by the paths used in the `#include` statements. E.g. pass `-Iinc` to the compiler, and then write:
> 
> ```c
> #include <a/api.h>
> #include <b/api.h>
> ```
> 
> In which case, it's pretty clear that you're actually _intentionally_ including both headers.
> 
> > Header guards don't have this problem because they work purely in the preprocessor's token space
> 
> That's actually exactly the problem with them. Because there's no guaranteed way to uniquely name them, it's quite likely that if `a` and `b` are unrelated projects, both `a/api.h` and `b/api.h` might choose _the same_ guard name, such as `API_H`. The more generic the name of the header, the more likely this is to happen.
> 
> In fact, this is exactly the scenario that led me to ask @antoinevg for this change. I was working on our Portapack firmware project that @mossmann mentioned in [portapack-mayhem/mayhem-firmware#2957 (comment)](https://github.com/portapack-mayhem/mayhem-firmware/issues/2957#issuecomment-3874193138), and I ended up needing to include both `gpio.h` from the HackRF source, and `gpio.hpp` from the PortaPack source. Both of them used the same guard name, `__GPIO_H__`, making it impossible to use both at once. At that point the only option was to modify one of the projects to prevent the clash.
> 
> This is where `#pragma once` is a better solution, precisely because it _doesn't_ depend on choosing unique names. It also avoids a lot of other mistakes that are common with traditional guards, like:
> 
> * clashes with system headers (officially, all identifiers starting with `__` are reserved to the C implementation!)
> * inconsistency about how to name guards within the project (e.g. we have a mix of `__NAME_H` and `__NAME_H__`)
> * forgetting to change a guard name when renaming a file, or copying code from one file to another
> * mismatching the `#ifndef` and the `#define`, (as I [fixed recently in our libopencm3 submodule](https://github.com/mossmann/libopencm3/pull/33)).
> 
> The main issue with `#pragma once` was that it wasn't always supported by all compilers, but that hasn't been the case for [a very long time now](https://en.wikipedia.org/wiki/Pragma_once#Portability) - it's been in GCC since 3.4 in 2004, in clang from the start, and in everything else for ages too by now.

No offenses intended. But, from the very same article you are mentioning, carefully read the caveats section. Maybe my case is not everyone's case, I'm working on legacies code bases on servers, with lots of symlinks and source files copies. I did not choose that, I inherited that and fix as I can. And pragma got our asses more than one time. I guess for hackRF code it's ok, but it's not the best move ever, from my point of view. I'm surely getting old, at least for code opinions and view points....

martinling (changes requested):
This is missing the same change in the recently-added `da7219.h`.

antoinevg:
> This is missing the same change in the recently-added `da7219.h`.

Whoopsie tx! Fixed and force-pushed.

martinling:
> No offenses intended.

None taken!

> But, from the very same article you are mentioning, carefully read the caveats section. Maybe my case is not everyone's case, I'm working on legacies code bases on servers, with lots of symlinks and source files copies. I did not choose that, I inherited that and fix as I can. And pragma got our asses more than one time. I guess for hackRF code it's ok, but it's not the best move ever, from my point of view.

I have read that, and the claim that `#pragma once` "can do the wrong thing" is marked with *[citation needed]*

Are you able to describe a specific scenario that may cause a problem - with compilers from the last 20 years?

I've been doing a bit of research about this, because it's been feeling to me like "`#pragma once` is unreliable" might have become one of those bits of folk wisdom that us older folks remember, but for reasons that are no longer quite so clear.

What I've found is that support for it was first added to GCC in 2001 in the venerable [2.95](https://gcc.gnu.org/gcc-2.95/features.html) release, but with a buggy implementation which _would_ sometimes get confused and include things twice. In 2003, it was deprecated and slated for removal, both due to that bug, and the fact it was not yet standardised.

However, the implementation was then fixed in 2004, released in version 3.4 and the deprecation removed.

I've yet to find anything definitive that suggests it's been an actual problem since!

martinling (approved):

PR #1705, opened 2026-03-24
Refactor `hackrf_core.x` in preparation for removal. (1/3)

antoinevg:
This PR moves a number of items out of `hackrf_core` in preparation for removing it entirely:

* CPU clock logic => `cpu_clock.x`
* Clock generator logic => `clock_gen.x`
* Power logic => `power.x`
* LED logic => `leds.x`
* `clkin_ctrl_set` => `clkin.x`
* `narrowband_filter_set` => `fpga.x`
* `trigger_enable` => `clock_gen.x`

Follow-up PR's address the remaining items:

* #1709
* #1710


firmware/common/radio.h
```diff
@@ -54,13 +54,6 @@ typedef struct {
 	uint32_t hz;
 } radio_sample_rate_t;
 
-// legacy type, moved from hackrf_core
-typedef enum {
-	CLOCK_SOURCE_HACKRF = 0,
-	CLOCK_SOURCE_EXTERNAL = 1,
-	CLOCK_SOURCE_PORTAPACK = 2,
-} clock_source_t;
```
mossmann, line 62: Looks like this is removed in the wrong commit.

firmware/common/hackrf_core.c
```diff
@@ -21,30 +21,27 @@
...
 #include <libopencm3/lpc43xx/scu.h>
 #include <libopencm3/lpc43xx/ssp.h>
-
-#include "platform_detect.h"
-
 #ifdef IS_NOT_RAD1O
 	#include <libopencm3/lpc43xx/ccu.h>
 #endif
 
-#include "delay.h"
```
mossmann, line 35: Removed too soon. ae0679bd does not compile.

firmware/common/hackrf_core.c
```diff
@@ -1152,34 +760,11 @@ void halt_and_flash(const uint32_t duration)
 	}
 }
 
-#ifdef IS_PRALINE
```
mossmann, line 1376: #ifdef removed too soon. ae0679bd does not compile.

firmware/common/hackrf_core.h
```diff
@@ -78,14 +78,7 @@ void pin_setup(void);
 
 void trigger_enable(const bool enable);
 
-#ifdef IS_PRALINE
```
mossmann, line 116: #ifdef removed too soon. 1868f238 does not compile.

firmware/common/fpga.c
```diff
@@ -137,3 +138,8 @@ void fpga_set_tx_nco_pstep(fpga_driver_t* const drv, const uint8_t phase_increme
 	set_FPGA_STANDARD_TX_PSTEP(drv, phase_increment);
 	fpga_regs_commit(drv);
 }
+
+void narrowband_filter_set(const uint8_t value)
```
mossmann: I don't like moving this to fpga.c because the narrowband filter is a distinct analog element that isn't logically associated with the FPGA. I think it would make the most sense to put it in rf_path.c or maybe in its own file.

firmware/common/clock_gen.c
```diff
@@ -427,6 +428,20 @@ fp_28_36_t sample_rate_set(const fp_28_36_t sample_rate, const bool program)
 	return resultant_rate;
 }
 
+void trigger_enable(const bool enable)
```
mossmann: I don't like this being in clock_gen.c as Harmony may want to include trigger I/O without including the clock generator. The only reason it is associated with the clock generator is that the trigger signals are accessible via the SMA clock connectors on one platform. Come to think of it, I'm not a big fan of moving those other GPIO controls here either. Maybe they could all go into clkin.c and rename it to clkio.c?

mossmann (changes requested):
Overall this looks good, and the final commit builds fine, but I would like to see some rearrangement.

firmware/common/fpga.c
antoinevg: There were linker problems with doing this before but it looks like #1715 resolved them. Yay!

firmware/common/clock_gen.c
antoinevg: 👍

mossmann (approved):
Nice, thank you!

PR #1709, opened 2026-03-31
Move `hackrf_core.x` driver instances to `drivers.x` (2/3)

antoinevg:
Depends on #1705 

---

### DIFF
--- a/firmware/common/clock_gen.c
+++ b/firmware/common/clock_gen.c
@@ -23,7 +23,7 @@
 
 #include <stdint.h>
 
-#include "hackrf_core.h"
+#include "drivers.h"
 #include "hackrf_ui.h"
 #include "i2c_bus.h"
 #include "platform_detect.h"
--- a/firmware/common/clock_io.c
+++ b/firmware/common/clock_io.c
@@ -30,9 +30,9 @@
 #include <libopencm3/lpc43xx/gpdma.h>
 #include <libopencm3/lpc43xx/creg.h>
 
+#include "drivers.h"
 #include "gpdma.h"
 #include "gpio.h"
-#include "hackrf_core.h"
 #ifdef IS_PRALINE
 	#include "fpga.h"
 	#include "platform_gpio.h"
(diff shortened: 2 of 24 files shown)


mossmann (approved):
I wonder if these would be better in hackrf_usb.c or distributed through various source files instead of continuing to lump them together into a common .c file, but I'll let @martinling make that decision later if this doesn't work for his effort. I think this is fine for now.

martinling:
For integration into PortaPack firmware, the following bits of `drivers.c` are problematic:

### `radio`

```c
radio_t radio = {
        .sample_rate_cb = sample_rate_set,
};
```

The reason for the `sample_rate_cb` field is to allow different clock generator logic to be used - in particular the existing `clock_manager.cpp` implementation in the PortaPack firmware, which has extra features. Setting the callback here in `drivers.c` adds an unnecessary dependency on our own `clock_gen.c`, even if the `sample_rate_cb` is changed at runtime.

For this reason, I think the `radio` instance definition should be somewhere under `firmware/hackrf_usb`.

That said, I'm working on making it possible for our `clock_gen.c` to replace `clock_manager.cpp` entirely, which seems like a better way forward given all the precision improvements made there recently. so this is probably a temporary issue.

### `spi_flash`

```c
w25q80bv_driver_t spi_flash = {
        .bus = &spi_bus_ssp0,
        .target_init = w25q80bv_target_init,
};
```

This adds an unnecessary dependency on `w25q80bv_target.c` which won't be used in a PortaPack build, since the SPI flash is kept in SPIFI mode there and we access the SPI flash via the memory mapping.

### Suggestion

I agreed with the decision to merge this, to advance the process of breaking up `hackrf_core`, but going forward, I think I'd be inclined to also break up `drivers.c`, and distribute the instances into the drivers themselves, e.g:

- `i2c0`, `i2c1` -> `i2c_bus.c`
- `ssp_config_max283x`, `max283x` -> `max283x.c`
- `clock_gen` -> `clock_gen.c`, or `si5351.c`
- ...etc

We can't use the instances without their drivers, and it's hard to see a scenario where we'd want the drivers without their instances.

PR #1710, opened 2026-03-31
Rename `hackrf_core.x` to `pins.x` (3/3)

antoinevg:
Depends on #1709

---


antoinevg:
I've put this back in draft for now because the `hackrf_core.x` rename operation got lost somewhere during rebase and it would be nice not to lose history.

mossmann:
If your concern is about GitHub representing the change to pins.x as a rename in "files changed" for the PR, I don't think you can control that detection. It is represented as a rename in the commit: [7cf005d](https://github.com/greatscottgadgets/hackrf/commit/7cf005d83b908a31137a6b615e84c55eb8b42a64)

antoinevg:
> If your concern is about GitHub representing the change to pins.x as a rename in "files changed" for the PR, I don't think you can control that detection. It is represented as a rename in the commit: [7cf005d](https://github.com/greatscottgadgets/hackrf/commit/7cf005d83b908a31137a6b615e84c55eb8b42a64)

Sometimes GitHub does a better job of reconstructing file "rename" history in non-commit views if you re-stage after a complicated rebase.  🤷

mossmann (approved):

PR #1730, opened 2026-04-11
Add macros to clarify fixed-point constants.

martinling:
It's not very clear at a glance what order of magnitude e.g. `21800000ULL * SR_FP_ONE_HZ` is, so add some macros for defining constants in each fixed-point format in Hz, kHz or MHz.


mossmann (approved):
Nice!

PR #1715, opened 2026-04-02
Unified Firmware Redux²

martinling:
This is a rebased version of:
- #1657 

plus some additional CI work to build and test `UNIVERSAL` firmware, and to simplify with macros.

PR #1734, opened 2026-04-16
HIL CI Fix

grvvy:


martinling (approved):
Looks good to me. I'll merge it, rebase #1715 and see what happens.

martinling:
Oh right it was set to merge into that branch. Seems all good then. :tada:

martinling:
@grvvy Since we were seeing this issue on `main` as well, I suggest we merge these changes to `main` in a new PR and then I'll rebase #1715 on top. That PR is already a huge change and I'd like to keep it tidy for the sake of future bisection etc.

I don't think the delay needs to be considered temporary - it's normal to have to wait a moment between `hackrf_spiflash` completing and seeing the rebooted device show up. I regularly have a `sleep 1` in there when testing things locally.

It makes sense that we see this problem on HackRF Pro more than HackRF One - the Pro takes longer to start up because it has to configure the FPGA (though I do have some ideas to make that go faster in future).

The fact that the device is taking longer to become available _might_ be due to `udev` rules on the server that are  slow to run for some reason - it may be worth having a quick audit of what's in `/etc/udev/rules.d` and remove anything unnecessary. This having changed since the server rebooted could be due to some change not being reloaded until that happened. But it's not really a big issue if it's slow - just set the delay long enough to cover the worst case.

PR #1742, opened 2026-04-28
experimental: Truly universal firmware

martinling:
With these changes, a `UNIVERSAL` build should also run on a Jawbreaker or rad1o badge.

```
Hardware supported by installed firmware:
    Jawbreaker
    rad1o
    HackRF One
    Praline
```


mossmann:
I added this to the next release milestone. It isn't necessarily required, but I want to resolve a naming concern before the release: If we are ever going to have a truly universal build, we probably don't want to establish the precedent that "universal" means only two platforms.

PR #1745, opened 2026-04-30
Firmware: Move driver instances into driver implementation files

antoinevg:
This PR moves driver instances out of a central `drivers.x` file and into the individual driver implementation files.

This shift is to reduce linker coupling that would result in unnecessary inclusion of symbols from `firmware/common/` in HackRF firmware other than `hackrf_usb`.

As per conversation at: https://github.com/greatscottgadgets/hackrf/pull/1709#issuecomment-4334443402


firmware/blinky/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ include(../hackrf-common.cmake)
 
 set(SRC_M4
 	blinky.c
+	"${PATH_HACKRF_FIRMWARE_COMMON}/cpld_jtag.c"
```
martinling: It's a bad sign if we're having to put anything here other than `blinky.c`, at least unless it contains symbols directly referenced by `blinky.c`.

The reason `cpld_jtag.c` ends up needed here is the `jtag_gpio_cpld` structure, which is required by  `pins.c`. Although we call `pins_setup` in blinky.c, `pins.c` is an automatic dependency via  `hackrf-common.cmake`, so I think `cpld_jtag.c` also belongs in `hackrf-common.cmake`, not here.

martinling:
From an integration point of view this is a good improvement and gets rid of most of the awkward legacy of `hackrf_core` being a module that depends on everything else in the tree.

Looking a bit beyond the scope of this PR, there is still a bit of a problem with `pins.c` depending on things that it doesn't really need to, because of code in `pins_setup` which writes to `ssp_config_max283x`, `ssp_config_max5864`, `sgpio_config`, `ssp_config_w25q80bv`, `spi_flash`, etc. Much of that initialization code doesn't need to be part of `pins_setup` - it could move into `init`/`setup` functions within the driver files that now own those structures.

firmware/blinky/CMakeLists.txt
antoinevg: Yeah, I wasn't too happy about that either. I agree it makes sense to put it in `hackrf-common.cmake`. I've pushed the change.

antoinevg:
> Looking a bit beyond the scope of this PR, there is still a bit of a problem with `pins.c` depending on things that it doesn't really need to, because of code in `pins_setup` which writes to `ssp_config_max283x`, `ssp_config_max5864`, `sgpio_config`, `ssp_config_w25q80bv`, `spi_flash`, etc. Much of that initialization code doesn't need to be part of `pins_setup` - it could move into `init`/`setup` functions within the driver files that now own those structures.

💯 -- this is one of the things I've been looking at in my carcinization branch. Tight-coupling is a blessing for readability when there is complex, inter-dependent configuration steps but most of `pins_setup()` really is just boilerplate that doesn't need to be right in your face.

martinling (approved):

mossmann (approved):
Nice! Thank you!

PR #1746, opened 2026-05-01
Improve pins_shutdown()

mossmann:
Pull down more pins in pins_shutdown() to reduce leakage current that can interfere with power sequencing and FPGA configuration on some Pralines.

Fix #1696


antoinevg (approved):
Tested on:

* `r1.0`: 👍
* `r1.2` - pre-production: 👍
* `r1.2` - production: 👍

Issue #1099, opened 2022-05-14
External clock not recongnized


gauxonz:
### Steps to reproduce
1. Power on my hackrf one without a TCXO then run `hackrf_debug --si5351c -n 0 -r`, I got `[  0] -> 0x50`
2. Plug a 10Mhz TCXO module and reboot the hackrf, run `hackrf_debug --si5351c -n 0 -r` again, the outcome stays the same  `[  0] -> 0x50`
![884873328](https://user-images.githubusercontent.com/2452969/168431937-97100c93-df0d-4155-a079-0db9be1f64a3.jpg)

### Expected behaviour
Accroding to the [wiki](https://hackrf.readthedocs.io/en/latest/external_clock_interface.html), the outcome of `hackrf_debug --si5351c -n 0 -r` should be either `0x51` or `0x01` (Is this correct ?)

### Actual behaviour
The result is always `0x50`, like below:
![image](https://user-images.githubusercontent.com/2452969/168432454-994fe604-a089-4c1c-83c3-2d39b3e3cfd5.png)

### Version information
**Operating system**:
Linux gz-PC-01 5.15.0-30-generic #31-Ubuntu SMP Thu May 5 10:00:34 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
**hackrf_info output:**
hackrf_info version: unknown
libhackrf version: unknown (0.6)
Found HackRF
Index: 0
Serial number: 0000000000000000723864c02e58a08f
Board ID Number: 2 (HackRF One)
Firmware Version: 2021.03.1 (API:1.04)
Part ID Number: 0xa000cb3c 0x00614761

simonboydfoley1:
Same issue it may be firmware related as it's supposed to read 0x51 not 0x50 if the TCXO is not detected. Perhaps things have changed With The latest firmware release. I checked the power and the External clock was getting 3.1v. So I will get a scope on the clock and make sure it's working as expected. Turns out mine was a clone sold by Amazon on eBay that was second hand and had had some reflow work done on it and the Design Fault on the Amp chip had occurred and it was blown too so I'm sending it back. I'm probably going to get a Pluto Plus because that design fault will just keep on popping the Amp as I have an internal VHF amp in my house to gain the signal to my shack room / office. I'll let you know if I get it working.

miek:
I saw a similar issue with another user recently, some clone HackRFs seem to be shipping without CLKIN support.

[...]

PR #1747, opened 2026-05-02
Si5351C refactoring

martinling:
This PR refactors the Si5351C driver code. The goal is to make it more maintainable and easier to add new features, by depending less on hand-written bit wrangling.

1. Although this PR is primarily an internal refactor, I've made a couple of changes to the external API used by `clock_gen.c` in order to better name and distinguish different concepts and types. These can be made independently of the refactoring work so are placed first.

2. Next, we add a new `si5351c_regs.def` file with register definitions and helper functions. This is similar to our other `def` files for other register-based parts, but with additional macros that simplify access to fields which are split across multiple registers, and for indexed access to fields which are duplicated for each multisynth, or for each PLL.

3. Finally, the implementation in `si5351.c` is rewritten to take advantage of the new helpers, while retaining the same structure and external API. As with other register-based parts, we now maintain a cached copy of the registers and a `regs_dirty` bitmask, and flush changes in a `si5351c_regs_commit()` function. The commit implementation will group writes to sequences of adjacent registers, so performance should match the previous hand-written code.

### DIFF
--- a/firmware/common/si5351c.h
+++ b/firmware/common/si5351c.h
@@ -64,11 +64,10 @@ typedef enum {
 
 #define SI5351C_CLK_PLL_SRC(x) ((x & SI5351C_PLL_B) << 4)
 
-enum pll_sources {
-	PLL_SOURCE_UNINITIALIZED = -1,
-	PLL_SOURCE_XTAL = 0,
-	PLL_SOURCE_CLKIN = 1,
-};
+typedef enum {
+	SI5351C_INPUT_XTAL = 0,
+	SI5351C_INPUT_CLKIN = 1,
+} si5351c_input_t;
 
 typedef struct {
 	i2c_bus_t* const bus;
@@ -80,12 +79,10 @@ void si5351c_disable_oeb_pin_control(si5351c_driver_t* const drv);
 void si5351c_power_down_all_clocks(si5351c_driver_t* const drv);
 void si5351c_set_crystal_configuration(si5351c_driver_t* const drv);
 void si5351c_enable_xo_and_ms_fanout(si5351c_driver_t* const drv);
-void si5351c_configure_pll_sources(
-	si5351c_driver_t* const drv,
-	const enum pll_sources source);
+void si5351c_configure_inputs(si5351c_driver_t* const drv, const si5351c_input_t input);
 void si5351c_configure_pll_multisynth(
 	si5351c_driver_t* const drv,
-	const enum pll_sources source);
+	const si5351c_input_t input);
 void si5351c_reset_pll(si5351c_driver_t* const drv, si5351c_pll_t pll);
 void si5351c_configure_multisynth(
 	si5351c_driver_t* const drv,
@@ -100,7 +97,7 @@ void si5351c_set_int_mode(
 	si5351c_driver_t* const drv,
 	const uint_fast8_t ms_number,
 	const uint_fast8_t on);
-void si5351c_set_clock_source(si5351c_driver_t* const drv, const enum pll_sources source);
+void si5351c_change_input(si5351c_driver_t* const drv, const si5351c_input_t input);
 bool si5351c_clkin_signal_valid(si5351c_driver_t* const drv);
 
 void si5351c_write_single(si5351c_driver_t* const drv, uint8_t reg, uint8_t val);
--- a/firmware/common/si5351c.h
+++ b/firmware/common/si5351c.h
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2022 Great Scott Gadgets <info@greatscottgadgets.com>
+ * Copyright 2012-2026 Great Scott Gadgets <info@greatscottgadgets.com>
  * Copyright 2012 Jared Boone <jared@sharebrained.com>
  *
  * This file is part of HackRF.
@@ -28,46 +28,61 @@ extern "C" {
 
 #include <stdbool.h>
 #include <stdint.h>
-#include <stddef.h>
 
 #include "i2c_bus.h"
 #include "i2c_lpc.h"
 
 #define SI_INTDIV(x) (x * 128 - 512)
 
-#define SI5351C_CLK_POWERDOWN (1 << 7)
-#define SI5351C_CLK_INT_MODE  (1 << 6)
-#define SI5351C_CLK_FRAC_MODE (0 << 6)
+#define SI5351C_CACHED_REGS 188
 
-#define SI5351C_CLK_INV (1 << 4)
+typedef enum {
+	SI5351C_MODE_FRAC = 0,
+	SI5351C_MODE_INT = 1,
+} si5351c_mode_t;
 
-#define SI5351C_CLK_SRC(x)              (x << 2)
-#define SI5351C_CLK_SRC_XTAL            0
-#define SI5351C_CLK_SRC_CLKIN           1
-#define SI5351C_CLK_SRC_MULTISYNTH_0_4  2
-#define SI5351C_CLK_SRC_MULTISYNTH_SELF 3
+typedef enum {
+	SI5351C_SRC_XTAL = 0,
+	SI5351C_SRC_CLKIN = 1,
+	SI5351C_SRC_MULTISYNTH_0_4 = 2,
+	SI5351C_SRC_MULTISYNTH_SELF = 3,
+} si5351c_src_t;
 
-#define SI5351C_CLK_IDRV(x)  (x << 0)
-#define SI5351C_CLK_IDRV_2MA 0
-#define SI5351C_CLK_IDRV_4MA 1
-#define SI5351C_CLK_IDRV_6MA 2
-#define SI5351C_CLK_IDRV_8MA 3
-
-#define SI5351C_LOS   (1 << 4)
-#define SI5351C_REVID 0x03
+typedef enum {
+	SI5351C_DRIVE_2MA = 0,
+	SI5351C_DRIVE_4MA = 1,
+	SI5351C_DRIVE_6MA = 2,
+	SI5351C_DRIVE_8MA = 3,
+} si5351c_drive_t;
 
 typedef enum {
 	SI5351C_PLL_A = 0,
 	SI5351C_PLL_B = 1,
 } si5351c_pll_t;
 
-#define SI5351C_CLK_PLL_SRC(x) ((x & SI5351C_PLL_B) << 5)
-
 typedef enum {
 	SI5351C_PLL_MASK_A = 1,
 	SI5351C_PLL_MASK_B = 2,
 	SI5351C_PLL_MASK_BOTH = 3,
 } si5351c_pll_mask_t;
+
+typedef enum {
+	SI5351C_XTAL_6PF = 1,
+	SI5351C_XTAL_8PF = 2,
+	SI5351C_XTAL_10PF = 3,
+} si5351c_xtal_t;
+
+typedef enum {
+	SI5351C_OUTPUT_ENABLE = 0,
+	SI5351C_OUTPUT_DISABLE = 1,
+} si5351c_output_t;
+
+typedef enum {
+	SI5351C_DIV_1 = 0,
+	SI5351C_DIV_2 = 1,
+	SI5351C_DIV_4 = 2,
+	SI5351C_DIV_8 = 3,
+} si5351c_div_t;
 
 typedef enum {
 	SI5351C_INPUT_XTAL = 0,
@@ -77,7 +92,18 @@ typedef enum {
 typedef struct {
 	i2c_bus_t* const bus;
 	uint8_t i2c_address;
+	uint8_t regs[SI5351C_CACHED_REGS];
+	uint32_t regs_dirty[(SI5351C_CACHED_REGS + 31) / 32];
 } si5351c_driver_t;
+
+typedef struct {
+	bool power_down;
+	si5351c_mode_t mode;
+	si5351c_pll_t pll;
+	si5351c_src_t source;
+	si5351c_drive_t drive;
+	bool invert;
+} si5351c_clk_ctrl_t;
 
 void si5351c_disable_all_outputs(si5351c_driver_t* const drv);
 void si5351c_disable_oeb_pin_control(si5351c_driver_t* const drv);
@@ -107,10 +133,6 @@ bool si5351c_clkin_signal_valid(si5351c_driver_t* const drv);
 
 void si5351c_write_single(si5351c_driver_t* const drv, uint8_t reg, uint8_t val);
 uint8_t si5351c_read_single(si5351c_driver_t* const drv, uint8_t reg);
-void si5351c_write(
-	si5351c_driver_t* const drv,
-	const uint8_t* const data,
-	const size_t data_count);
 void si5351c_clkout_enable(si5351c_driver_t* const drv, uint8_t enable);
 void si5351c_init(si5351c_driver_t* const drv);
 void si5351c_set_phase(
--- a/firmware/common/si5351c.h
+++ b/firmware/common/si5351c.h
@@ -57,12 +57,17 @@ extern "C" {
 #define SI5351C_REVID 0x03
 
 typedef enum {
-	SI5351C_PLL_A = 1,
-	SI5351C_PLL_B = 2,
-	SI5351C_PLL_BOTH = 3,
+	SI5351C_PLL_A = 0,
+	SI5351C_PLL_B = 1,
 } si5351c_pll_t;
 
-#define SI5351C_CLK_PLL_SRC(x) ((x & SI5351C_PLL_B) << 4)
+#define SI5351C_CLK_PLL_SRC(x) ((x & SI5351C_PLL_B) << 5)
+
+typedef enum {
+	SI5351C_PLL_MASK_A = 1,
+	SI5351C_PLL_MASK_B = 2,
+	SI5351C_PLL_MASK_BOTH = 3,
+} si5351c_pll_mask_t;
 
 typedef enum {
 	SI5351C_INPUT_XTAL = 0,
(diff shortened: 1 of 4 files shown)


martinling:
This has been tested on HackRF One and Praline in the HITL CI, and also verified locally by taking diffs of `hackrf_debug` register dumps before and after the changes.

There is one remaining difference in my register dumps:

```
 [180] -> 0x00
 [181] -> 0x30
 [182] -> 0x1f
-[183] -> 0x80
+[183] -> 0x92
 [184] -> 0x60
 [185] -> 0x60
 [186] -> 0xb8
```

Previously, we were writing `0x80` to register 183, which sets the crystal load capacitance in the top two bits, and sets the rest of the register to zero. However, the documentation specifies that the reserved lower bits should be written as `010010`:

image

The new implementation does this, because it reads the reserved bits during setup and retains them when updating the `XTAL_CL` field later.

martinling:
There is another difference in register dumps after this change, which happens specifically on Praline:
```
 [ 16] -> 0x0d
 [ 17] -> 0x0c
 [ 18] -> 0x0c
-[ 19] -> 0xc0
+[ 19] -> 0xe0
 [ 20] -> 0x7d
 [ 21] -> 0x6d
 [ 22] -> 0xc0
```

This corresponds to `MS3_SRC` changing from 0 to 1, meaning that CLKOUT is now using PLL B instead of PLL A.

This is because of what I assume is a bug in the `IS_PRALINE` branch in `si5351c_configure_clock_control`, which previously did:
```c
/* CLK3: CLKOUT */
clkout_ctrl = SI5351C_CLK_INT_MODE | SI5351C_CLK_PLL_SRC(SI5351C_PLL_B)  |
        SI5351C_CLK_SRC(SI5351C_CLK_SRC_MULTISYNTH_SELF) |
        SI5351C_CLK_IDRV(SI5351C_CLK_IDRV_8MA);
```
...but never copied the new value of `clkout_ctrl` to the `data` array which was about to be sent over I2C. So the change to `clkout_ctrl`, which differed only in the PLL selection, never took effect.

I have assumed that the intent was to use PLL B for CLKOUT on Praline, so I now have:

```c
/* CLK3: CLKOUT */
clkout.pll = SI5351C_PLL_B;
clk[3] = clkout;
```

martinling:
Something is still wrong with this on HackRF One r9; achieved sample rate appears to be around 14MHz (with XTAL) or 15MHz (with CLKIN) when it should be 10MHz.

After startup and a `hackrf_transfer -r /dev/null` run, I get the following register differences before and after this PR:

```
@@ -1,4 +1,4 @@
-[  0] -> 0x11
+[  0] -> 0x71
 [  1] -> 0xf8
 [  2] -> 0x03
 [  3] -> 0xfc
@@ -17,9 +17,9 @@
 [ 16] -> 0x4e
 [ 17] -> 0x0d
 [ 18] -> 0xc0
-[ 19] -> 0x80
-[ 20] -> 0x80
-[ 21] -> 0x80
+[ 19] -> 0xc0
+[ 20] -> 0xc0
+[ 21] -> 0xc0
 [ 22] -> 0xc0
 [ 23] -> 0xc0
 [ 24] -> 0x00
@@ -220,8 +220,8 @@
 [219] -> 0x00
 [220] -> 0x00
 [221] -> 0x00
-[222] -> 0x00
-[223] -> 0x40
+[222] -> 0x94
+[223] -> 0x28
 [224] -> 0x00
 [225] -> 0x00
 [226] -> 0x38
```

Register 0 is the status register, and is indicating loss of lock on both PLLs, which would certainly explain the wrong sample rate, but I don't know why.

Registers 19-21 should be irrelevant; these are the clock control registers for the CLK3-CLK5 outputs, which don't exist on the Si5351A, and are also flagged as powered down.

Registers 222-223 are undocumented, and listed as reserved in AN619, so I've no idea what that change means.

martinling:
Found the problem on r9 - just a dumb mistake in the first commit, when updating the logic controlling `h1r9_clkin_en`.

mossmann:
> Previously, we were writing `0x80` to register 183, which sets the crystal load capacitance in the top two bits, and sets the rest of the register to zero. However, the documentation specifies that the reserved lower bits should be written as `010010`:

I did some testing and found that some Si5351Cs installed on H1s have those low bits set to 0 at start-up. I also found that changing those bits has no effect on clock frequency.

firmware/common/si5351c.h
```diff
@@ -57,13 +57,18 @@ extern "C" {
...
 
 typedef enum {
-	SI5351C_PLL_A = 1,
-	SI5351C_PLL_B = 2,
-	SI5351C_PLL_BOTH = 3,
+	SI5351C_PLL_A = 0,
+	SI5351C_PLL_B = 1,
 } si5351c_pll_t;
 
 #define SI5351C_CLK_PLL_SRC(x) ((x & SI5351C_PLL_B) << 4)
```
mossmann: `SI5351C_CLK_PLL_SRC` is broken here but gets removed in a later commit anyway. It can be fixed by changing 4 to 5.

mossmann (changes requested):
Great overall! I found one small bug.

firmware/common/si5351c.h
martinling: Oh, nice catch. Fixed in the commit that broke it, to keep a bisect-happy history.

mossmann (approved):

Brumi-2021:
Hi @martinling and @mossmann, first of all, congratulations on your great work across the GSG projects and on the excellent documentation.

In relation to this PR, I would like to share some investigation results from the Mayhem/PortaPack side, after observing a HackRF USB mode regression on one out of four HackRF devices I tested.

Apologies for the long post, but I wanted to include the experimental details and hardware results.

While investigating a HackRF USB mode regression affecting one particular HackRF One r6 + PortaPack H2 (WM8731), we eventually traced the issue back to the Si5351C changes introduced by this PR.

The affected unit works correctly as a PortaPack, but in HackRF USB mode
with applications such as SDR++ and GQRX it exhibits incorrect RF
reception: USB communication and IQ streaming continue to work, tuning
commands are accepted, and LNA/VGA/RF AMP controls work, but the received
spectrum is incorrect/mistuned and the expected RF signal cannot be
received correctly.

The investigation originally started as this Mayhem issue:

https://github.com/portapack-mayhem/mayhem-firmware/issues/3268

After extensive hardware testing and manual bisecting, we narrowed the
regression boundary to these two consecutive Mayhem nightlies:

    nightly 2026-05-21 -> GOOD
    nightly 2026-05-22 -> BAD

on the affected device.

The corresponding HackRF submodule revisions are:

    May21 GOOD:
      Mayhem: 865bdee9b37dcb5ac344db242035645f4bd1bc06
      HackRF: 0bf478ed259410a5cef085b6a50c795a80c096ad

    May22 BAD:
      Mayhem: f3aee44e3b9a09ec41bb46c8135fe85eae4b9562
      HackRF: f526ddf4ebba5fda5ad0075c78238ea0ce1cc02a

After isolating the HackRF changes in that interval, the regression was
further narrowed to commit:

    016910567bc236fcfb14ebb741185c468416cda8

from this PR.

More specifically, we were eventually able to isolate the physical
regression to the handling of Si5351 register 183.

Before this change, HackRF programmed REG183 deterministically as:

    REG183 = 0x80

The refactored driver first snapshots the Si5351 register map and uses
cached read/modify/write operations. For REG183, set_XTAL_CL() changes
D7:D6 while preserving D5:D0 from the cached value.

During the investigation we noticed the earlier comment in this PR about
REG183:

> Previously, we were writing 0x80 to register 183 [...]

and, in particular, the observation that some Si5351Cs installed on H1s
had the low bits set to zero at startup and that changing those bits did
not appear to affect clock frequency.

Our hardware testing has produced a different and reproducible result on
one r6/H2 unit.

Skyworks AN619 Rev. 0.8 specifies REG183 as:

    D7:D6 = XTAL_CL[1:0]
    D5:D0 = Reserved

and states that D5:D0 should be written as:

    010010b = 0x12

Therefore, with the existing 8 pF XTAL_CL setting:

    D7:D6 = 10b
    D5:D0 = 010010b

giving:

    REG183 = 0x92

This also matches what we observe in current HackRF firmware:

    hackrf_debug --si5351c -n 183 -r
    [183] -> 0x92

However, on the affected r6/H2 device, REG183=0x92 reproducibly causes
the RF failure, while REG183=0x80 reproducibly restores correct RF
operation.

We isolated this experimentally while retaining the new register
cache/dirty/commit architecture.

Historical isolated tests on the affected device:

    REG183 low bits preserved -> BAD-STREAMING-RF-WRONG
    REG183 forced to 0x80     -> GOOD
    REG183 forced to 0x92     -> BAD-STREAMING-RF-WRONG

The 0x80 and 0x92 test candidates used the same helper/cache/dirty/commit
path, the same initialization position and the same I2C transaction.
Only REG183 D5:D0 differed.

We then repeated the experiment on current Mayhem next / current HackRF.

Current upstream base:

    Mayhem:
    ca9ca93fa29c745d6550c6ad41c58479e5a8963a

    HackRF:
    749bca5f10c491906fbb1acc6fb466784b73ef43

Current next is BAD on the affected device and reports:

    [183] -> 0x92

We then made a one-line diagnostic change in the current Si5351 driver:

    void si5351c_set_crystal_configuration(si5351c_driver_t* const drv)
    {
        set_XTAL_CL(drv, SI5351C_XTAL_8PF);
        drv->regs[183] &= 0xC0;
        si5351c_regs_commit(drv);
    }

This keeps the current register cache/dirty/commit implementation intact
and only makes D5:D0 deterministic zero again, producing exactly:

    REG183 = 0x80

No I2C timing, clock source, PLL, multisynth, startup sequence or other
Si5351 configuration was changed.

The corresponding current-next test revisions are:

    Mayhem:
    1e55ade8896c76a8e8aa20eb2cbedfe919941c1f

    HackRF:
    1b7774135280ed2c73215e090a7921324d7fea8f

The firmware was built using the normal Mayhem Docker toolchain.

SHA256 of the tested HackRF One firmware:

    8c354a79c4aa0acc0de2cfd33ff834134992ceeac61b43a19ff63e75d90b4b2c

SHA256 of the tested HackRF Pro firmware:

    05f4703a9fe6212f3a50ea25597035c19b857e8a633222cec688425795add5c1

On the affected device this immediately changes the result from:

    current next + REG183 0x92 -> BAD
    current next + REG183 0x80 -> GOOD

We then tested the same current-next 0x80 candidate across four different
physical devices:

    Device 1:
      HackRF One r6 + PortaPack H2 + WM8731
      Previously affected
      REG183 = 0x80
      Result: GOOD

    Device 2:
      HackRF One r6 + PortaPack H1 + AK4951K
      Previously unaffected
      REG183 = 0x80
      Result: GOOD

    Device 3:
      HackRF One r9 + PortaPack H2 + WM8731
      Previously unaffected
      REG183 = 0x80
      Result: GOOD

    Device 4:
      HackRF Pro + PortaPack H4M + WM8731
      Previously unaffected
      REG183 = 0x80
      Result: GOOD

In all four cases the programmed value was also verified directly using:

    hackrf_debug --si5351c -n 183 -r
    [183] -> 0x80

So the current-next validation result is 4/4 GOOD with deterministic
REG183=0x80.

For comparison, REG183=0x92 is not generally broken: it works correctly
on the other tested hardware. For example, the r6/H1 and HackRF Pro both
operate correctly with 0x92.

Therefore we are NOT suggesting that AN619 is generally wrong, nor that
0x92 is generally invalid.

What we have found is at least one reproducible hardware case where:

    REG183 = 0x92 -> incorrect RF operation
    REG183 = 0x80 -> correct RF operation

while USB communication and IQ streaming themselves remain operational.

This particular affected HackRF One r6/H2 is an older third-party unit,
so a silicon/BOM/device variation is certainly possible. We have not
identified the underlying electrical reason why D5:D0 affect this unit
while apparently being benign on the others.

For reference, here is a photo of the Si5351 device fitted on the affected HackRF One r6/H2 PCB. The package marking is visible in the photo.
image

Given the earlier testing mentioned in this PR, we would be very
interested in your interpretation:

Could there be a known Si5351C silicon/BOM/compatible-device variation,
or an interaction with the 25 MHz reference implementation, that could
explain why REG183=0x92 is benign on most tested units but reproducibly
breaks RF operation on this particular r6/H2, while restoring the
historical REG183=0x80 behavior fixes it?

We can provide the individual test commits / firmware hashes and further
hardware measurements if useful.

miek:
One possible difference comes to mind: during the chip shortages a few years ago some third-party vendors decided to substitute in Si5351A instead of Si5351C on the boards they produced. Unfortunately there's no way to tell from the package marking which variant is fitted (the 'A' is for the chip revision not the model variant). 

However, since the Si5351A has no CLKIN support, they also added an extra footprint breaking the CLKIN PCB trace, so you could check to see if that is present. There are some pictures of that on issue #1099 . I do notice your board has the same marking & date code as the one in that issue, so I think it's pretty likely it's one of those boards.

Brumi-2021:
Thanks @miek, that's a very interesting observation.
Looking again at my PCB photo, the additional CLKIN footprint shown in #1099 appears to be present in approximately the same location, although unfortunately it is partially hidden by the P16 connector, so I cannot confirm its exact population/connection without reopening the unit.

image


I would prefer not to reopen it at the moment because this old board has some damaged solder-mask areas that I have had to insulate carefully.

Your observation about the identical package marking/date code is particularly interesting. This could indeed mean that the affected r6 board belongs to that batch of third-party boards with a possible Si5351A substitution.

One additional data point is that **my Device 3 is a HackRF One r9, therefore using the Si5351A design, and it works correctly with both REG183=0x92 and our 0x80 test.** So Si5351A alone would not seem sufficient to explain the REG183 sensitivity, but perhaps the combination of the substituted device and this particular r6/third-party board design is relevant.

image


</div></div>
<p data-start="851" data-end="873">From REG0, this gives:</p>
<p data-start="878" data-end="997"><code data-start="878" data-end="904">Device 1: REVID[1:0] = 0</code><br data-start="904" data-end="907">
<code data-start="909" data-end="935">Device 2: REVID[1:0] = 1</code><br data-start="935" data-end="938">
<code data-start="940" data-end="966">Device 3: REVID[1:0] = 1</code><br data-start="966" data-end="969">
<code data-start="971" data-end="997">Device 4: REVID[1:0] = 1</code></p>
<p data-start="1002" data-end="1316">I understand that <code data-start="1020" data-end="1027">REVID</code> identifies the silicon revision rather than the Si5351 A/B/C variant, so I am <strong data-start="1106" data-end="1113">not</strong> using this as evidence that Device 1 contains an Si5351A. However, it may be another useful clue: the affected board is also the only device in my four-unit sample reporting a different Si5351 revision.</p>
<p data-start="1321" data-end="1504">Also, Device 3 (known Si5351A design) and Device 4 (Si5351C) both return exactly <code data-start="1402" data-end="1415">REG0 = 0x11</code>, which seems to confirm that REG0 itself cannot be used to distinguish the A/C variants.</p>

martinling:
We don't know what the reserved bits of register 183 are. However, in general the Si5351 register map is very logical and related settings are placed together, so I suspect that those bits also relate to the crystal oscillator circuit somehow. They could be adjusting drive strength of the output for the driven side of the crystal, for instance.

It would be interesting to probe pin 2 (XB) of the Si5351 on the troublesome board to see how the crystal waveform differs between the 0x80 and 0x92 settings.

Brumi-2021:
Thanks @martinling , that is a very interesting possibility.

The idea that REG183 D5:D0 may control some undocumented analogue parameter of the oscillator block would fit the reproducible 0x80 / 0x92 behaviour we are seeing.

I will try tomorrow to probe XB (pin 2) on the affected board and compare the waveform with REG183=0x80 and REG183=0x92.

One detail worth mentioning is that Device 1 appears to use a passive 4-pad 25 MHz crystal, so I will first verify the exact XA/XB connection before interpreting the measurement.

In any case, experimentally we have already established that this particular Device 1 cannot operate correctly with external SDR programs  (sdr++, gqrx ,....)  with REG183=0x92, while restoring the previous REG183=0x80 value makes it work correctly and reproducibly.

Once we have the oscilloscope measurements, I would therefore appreciate considering whether some form of backward-compatibility detection could be appropriate for these older/special hardware variants, rather than changing the current behaviour for all devices.

As a preliminary example only, one possibility might be to use the Si5351 REVID value: the affected device is the only one in my four-device sample reporting REVID=0, while all three unaffected devices report REVID=1. This could potentially allow something like 0x80 for REVID=0 devices and the current AN619 0x92 behaviour for REVID>=1.

Of course, at this point we only have one REVID=0 device, so I would not assume that this is the correct detection criterion without further evidence. I mention it only as an example of the kind of backward-compatible solution that might be possible.

martinling:
It's worth noting that those REVID=0 parts are very old.

Even [this photo of a Jawbreaker board from 2012](https://hackrf.readthedocs.io/en/latest/jawbreaker.html) already shows an Si5351 with the `B` marking, i.e. REVID=1, so presumably the REVID=0 parts could have been manufactured no later than mid-2012. They must have lurked on a shelf for a decade until the 2021-2022 chip shortage, when those clones with the broken CLKIN appeared.

The Si5351 family appear to have been first marketed in late 2010, so it may be that REVID=0 are just the parts from the first year or two of production - but it's also possible that Rev. 0 was for pre-production samples, in which case the advice in AN619 might not even be intended to apply to them.

gullradriel:
Yeah, some of our users (including some of us maintainers) are having old devices. I don't see a nice way to tell it, but we are trying to keep compatibility even with old devices.

martinling:
I've opened PR #1811, which implements the proposal to zero the reserved bits of register 183 for rev 0 parts, whilst following the advice in AN619 otherwise.

It's possible that the reason this helps is not related to the Si5351 revision, but rather to something about the crystal used on that specific clone board. However, even if that's the case, this could still be the right solution. Since the rev 0 chips predate the entire HackRF project, it's quite possible that this one batch of chip-shortage-era clones, with Si5351A and broken CLKIN, are the only HackRF boards that will ever exist with rev 0 chips on them.

What's still not clear is whether 0x80 is the _best_ choice of register value for those boards  - perhaps the crystal has a different load capacitance, or perhaps zero is not the best value for the mystery bits. You could investigate that if you want to, but 0x80 seems to work, so if you're happy with that we can go with it.

PR #1749, opened 2026-05-07
libhackrf: Set timeouts for all USB transfers

martinling:
Currently, libhackrf makes many USB control requests with no timeout set. If those requests fail to complete for any reason other than a device disconnect, libhackrf will hang indefinitely.

This PR sets timeouts for all USB transfers, with a default of 100ms for most USB requests, and longer times for CPLD writes (10 seconds, unchanged) and SPI flash erase/writes (50 seconds, per W25Q32JV datasheet).


mossmann (approved):
I wouldn't be surprised if this shakes out some bugs, and we end up increasing the default timeout and/or adding some error handling, but I think that would be preferable to the current behavior.

PR #1751, opened 2026-05-09
Further work on Si5351 driver

martinling:
Further work on the Si5351 driver:

- Wait for the `SYS_INIT` bit to clear in the status register before proceeding with setup.
- Make more use of the `clk` array to clarify code.
- Add control of `MCU_CLKIN` output.

### DIFF
--- a/firmware/common/si5351c.c
+++ b/firmware/common/si5351c.c
@@ -51,12 +51,6 @@ si5351c_driver_t si5351c = {
 	.i2c_address = 0x60,
 };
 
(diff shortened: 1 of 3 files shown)


martinling:
I just realised we ought not to be using globals for state in `si5351c.c` when we have a driver struct for that, so have moved things into the driver struct.

mndza (approved):

mossmann (approved):
Nice! Thank you!

PR #1753, opened 2026-05-13
Reduce dependencies of `pins.c`

martinling:
Moves various driver-specific code out of `pins.c` and into the `setup` and `shutdown` functions of the relevant drivers, so that it's possible to build `pins.c` without depending on all the drivers.

### DIFF
--- a/firmware/common/max283x.c
+++ b/firmware/common/max283x.c
@@ -26,6 +26,8 @@
 #include <stdbool.h>
 #include <string.h>
 
+#include <libopencm3/lpc43xx/ssp.h>
+
 #include "fixed_point.h"
 #include "platform_detect.h"
 #include "platform_gpio.h"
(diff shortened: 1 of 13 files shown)


firmware/common/max283x.c
```diff
@@ -58,6 +60,20 @@ max283x_driver_t max283x = {};
...
+#ifdef IS_NOT_PRALINE
+	if (IS_NOT_PRALINE) {
+		ssp_config_max283x.data_bits = SSP_DATA_16BITS;
+	}
+#endif
+#ifdef IS_PRALINE
+	if (IS_PRALINE) {
+		ssp_config_max283x.data_bits = SSP_DATA_9BITS; // send 2 words
+	}
+#endif
```
antoinevg: This was probably meant to go into `max283x_setup()`?

antoinevg (changes requested):

firmware/common/max283x.c
martinling: You would think so, but I tried that, and things break if it's put there.

Presumably because something is calling `ssp1_set_mode_max283x` before `max283x_setup` gets run.

martinling: Fixed by rebasing on #1765, which is ready for your review.

martinling:
@antoinevg this needed a bit more fixing but should now be good to go.

antoinevg (approved):
👍

Tested on HackRF One r4 and r9, Praline r1.0 and r1.2 with both UNIVERSAL and board targets.

PR #1755, opened 2026-05-18
Add script to extract font data to an image file for editing.

martinling:
Adapted from a [similar script](https://github.com/martinling/UHSDR/blob/active-devel/mchf-eclipse/drivers/ui/lcd/edit-8x8-font.py) I wrote for the UHSDR firmware project.


mossmann (approved):
Handy!

martinling:
I used this plus some manual fiddling to make the font changes in #1756, but it's still fairly fragile (in particular, it assumes the number of characters will be divisible by 5, which after that PR is no longer true). I guess it's still worth having as-is though.

PR #1759, opened 2026-05-19
Move LPC43xx I2C instances and config to `i2c_lpc.{c,h}`

martinling:
It makes more sense to put `i2c0` and `i2c1` in `i2c_lpc`, since they're the I2C resources available on the LPC43xx, whereas `i2c_bus` is a generic middleware layer.

Similarly `i2c_config_fast_clock` is an LPC43xx I2C config for fast I2C (400kHz) at the usual input clock rate of 204MHz, so it doesn't belong specifically in the `si5351c` driver.

### DIFF
--- a/firmware/common/clock_gen.c
+++ b/firmware/common/clock_gen.c
@@ -25,6 +25,7 @@
 
 #include "hackrf_ui.h"
 #include "i2c_bus.h"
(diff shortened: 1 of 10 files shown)


antoinevg (approved):
👌

PR #1762, opened 2026-05-22
Fix reads with hackrf_spiflash when size is a multiple of 64 bytes.

### LINKED ISSUE #1659: Reading flash with `hackrf_spiflash` fails if size is a multiple of 64 bytes.

This broke in #1613 - see https://github.com/greatscottgadgets/hackrf/pull/1613#issuecomment-3801654648

However, that PR fixed other cases, like retrieving the self-test report.

`hackrf_spiflash -r flash.bin -l 256`

See packet captures in https://github.com/greatscottgadgets/hackrf/pull/1613#issuecomment-3801654648.

mndza:
Fix #1659.

### DIFF
--- a/firmware/hackrf_usb/usb_api_spiflash.c
+++ b/firmware/hackrf_usb/usb_api_spiflash.c
@@ -111,6 +111,7 @@ usb_request_status_t usb_vendor_request_read_spiflash(
 				len,
 				NULL,
 				NULL);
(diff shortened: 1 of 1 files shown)


antoinevg (approved):
🏅 For conspicuous gallantry under hostile USB conditions, and distinguished service to Endpoint Zero, this ribbon is awarded for the successful identification and deployment of the missing ZLP.

PR #1765, opened 2026-05-29
Automatic SPI bus reconfiguration

martinling:
This eliminates the `ssp1_set_mode_*` calls scattered around the code.

The key change is that `spi_transfer` now takes a `config` argument. If the bus is already set up for that config, nothing happens. If it is not, the relevant `spi_bus_start` call is made automatically.

Additionally, this PR moves `spi_bus_ssp0` and `spi_bus_ssp1` out of `spi_bus` and into `spi_ssp`, for the same reasons as #1759.

### DIFF
--- a/firmware/common/ui_rad1o.c
+++ b/firmware/common/ui_rad1o.c
@@ -26,7 +26,6 @@
 #include <stdint.h>
 
 #include "clock_gen.h"
-#include "max283x.h"
 #include "rf_path.h"
 #include "transceiver_mode.h"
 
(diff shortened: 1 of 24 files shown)


firmware/common/ui_rad1o.c
```diff
@@ -195,9 +194,6 @@ static void ui_update(void)
 	rad1o_lcdNl();
 
 	rad1o_lcdDisplay();
-
-	// Don't ask...
```
antoinevg, line 199: 🥲

antoinevg (approved):
This is great.

Checked on HackRF One `r4` and `r9`, Praline `r1.0` and `r1.2` with both `UNIVERSAL` and board targets.

PR #1766, opened 2026-05-29
Clean up delay functions

martinling:
We had a rather scattered selection of delay functions being used:

- `delay`  simply took a number of loop iterations, for a C loop with a `nop` in it. Analysis of this loop as compiled by GCC 15.2 shows it currently takes 7 cycles per iteration on the M4. This function has been removed, and all uses of it have been replaced with calls to `delay_us` or `delay_ms` according to either apparent intent, or actual effect, at each call site.
```
00004618 <delay>:
    4618:       2300            movs    r3, #0
    461a:       4283            cmp     r3, r0           // 1 cycle
    461c:       d100            bne.n   4620 <delay+0x8> // 2 cycles, branch taken
    461e:       4770            bx      lr               // skipped
    4620:       bf00            nop                      // 1 cycle
    4622:       3301            adds    r3, #1           // 1 cycle
    4624:       e7f9            b.n     461a <delay+0x2> // 2 cycles, branch taken

```
- `delayms` in `rad1o/display.c`, which builds on `delay` but "_naively assumes 204 MHz instruction cycle clock and five instructions per count_" (wrong, though might have been right on earlier compilers). Replaced with `delay_ms`.
- `portapack_sleep_milliseconds`, which did the same. Replaced with `delay_ms`.
- `delay_us_at_mhz`, which I added to provide an actually-calibrated loop, but which requires knowing the current CPU clock speed. Replaced with `delay_us`, which uses `delay_us_at_mhz` internally, but with the clock speed parameter provided by the new `cpu_clock_mhz` symbol, updated by code in `cpu_clock` when making changes.
- `halt_and_flash` passed its argument to `delay`. Replaced that call with `delay_ms`, and updated arguments to `halt_and_flash` accordingly.

### DIFF
--- a/firmware/blinky/blinky.c
+++ b/firmware/blinky/blinky.c
@@ -45,13 +45,13 @@ int main(void)
 		led_on(LED2);
 		led_on(LED3);
 
-		delay(2000000);
+		delay_ms(150);
 
 		led_off(LED1);
 		led_off(LED2);
 		led_off(LED3);
 
-		delay(2000000);
+		delay_ms(150);
 	}
 
 	return 0;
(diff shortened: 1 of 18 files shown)


mndza (approved):
Great!

PR #1768, opened 2026-05-29
firmware: safe CPLD SRAM-read checksum and FPGA bitstream guard

galic1987:
## Summary

This PR addresses three long-standing issues:
1. **CPLD checksum crashes** — Replaces fragile flash-read path with safe SRAM-read (~3ms, deterministic cleanup)
2. **FPGA bitstream corruption risk** — Adds guard against active streaming by checking applied radio opmode
3. **Host timeout too short** — Increases FPGA bitstream timeout from 100ms to 1000ms

## Changes

- : Add  (SRAM-read, ~3ms)
- : Use SRAM-read, remove dead error branch
- : Guard  with 
- : Add , add  API
- : Add  /  flag
- : Fix  /  optstring conflict

## Testing

- [x] HackRF One r9 + PortaPack H2: CPLD checksum returns 
- [x] HackRF Pro r1.2: FPGA switch succeeds when idle, STALLs during RX
- [x] Both devices: Usage:
	-h, --help: this help
	-a, --address <n>: starting address (default: 0)
	-l, --length <n>: number of bytes to read (default: all)
	-r, --read <filename>: Read data into file.
	-w, --write <filename>: Write data from file.
	-i, --no-check: Skip check for firmware compatibility with target device.
	-d, --device <serialnumber>: Serial number of device, if multiple devices
	-s, --status: Read SPI flash status registers before other operations.
	-c, --clear: Clear SPI flash status registers before other operations.
	-R, --reset: Reset HackRF after other operations.
	-v, --verbose: Verbose output. clears status correctly

## API Changes

- New:  — requires API 0x0103
- Modified:  timeout increased to 1000ms

### DIFF
--- a/ACHIEVEMENTS_REPORT.md
+++ b/ACHIEVEMENTS_REPORT.md
@@ -1,0 +1,510 @@
+# HackRF Firmware & Hardware Development Session — Technical Achievement Report
+
+**Date:** 2026-05-28  
+**Scope:** Firmware (`firmware/`), Host Library (`host/libhackrf/`), Tools (`host/hackrf-tools/`, `ci-scripts/`), Hardware Validation  
+**Target Platforms:** HackRF One r9, HackRF Pro r1.2 (Praline), PortaPack H2  
+**Status:** Validated on hardware, ready for upstream contribution review  
+
+---
+
+## 1. Executive Summary
+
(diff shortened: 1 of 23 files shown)


mossmann:
We do not accept LLM-generated code. If any of the concerns addressed by these PRs is something that has caused an actual problem for you, we would appreciate it if you would report them as individual issues.

PR #1771, opened 2026-06-01
Add support for locking radio registers (1/n)

antoinevg:
Depends on: #1806 

---

This PR adds support for locking radio registers.

```
Usage:

    # Lock the RADIO_FREQUENCY_RF register
    hackrf_debug --radio --register 1 --lock 1
    
    # Unlock the RADIO_FREQUENCY_RF register
    hackrf_debug --radio --register 1 --lock 0
```

Implementation decisions:

* We don't have multi-field registers so locks are coarse-grained over the entire register.
* Attempting to write to a locked register returns an error: `RADIO_ERR_LOCKED_REGISTER`
* A lock prevents changing the register value rather than the application of a register value. This is to avoid a situation where registers are locked, other host software makes unanticipated changes and then there is a potentially unknown value in the register when it is unlocked.
* Locks apply independent of the selected bank rather than per-bank. i.e. You can't have a register which is locked for `RX` but unlocked for `APPLIED`, `ACTIVE`, `IDLE`, `TX` and `ALL`.
* Calling `radio_init()` clears all register locks.


martinling:
@miek and I were just discussing this but we don't really understand the use case yet - can you explain a bit?

PR #1781, opened 2026-06-12
Simplify PortaPack detection and init API

martinling:
Previously, we had a somewhat overcomplicated method for PortaPack detection and conditional usage.

The `portapack` weak symbol was used to detect both whether PortaPack support was compiled-in (i.e. not `RAD1O` or `JAWBREAKER` builds), and if not `NULL`, could be called to detect whether a PortaPack was present at runtime, returning either a pointer to a  structure or `NULL`.

There was another weak symbol for `portapack_reference_oscillator` which could also be `NULL`.

Most of this logic has been obsoleted by our new platform logic in #1715, and at this point all we need is a simple boolean test for whether a PortaPack is present.

This PR provides that in the form of `portapack_present()`, and for convenience, also returns this from `portapack_init()`.


antoinevg (approved):
👍

PR #1788, opened 2026-06-24
si5351c: Add support for compensating clock reference error

martinling:
Work in progress.

Adds support for specifying an error in ppm for the reference oscillator input (either crystal or CLKIN), and adjusts the Si5351C PLL multisynth programming accordingly such that downstream frequencies are correct.


martinling:
Abandoning this approach as it leads to higher phase noise. Instead, we will add a radio register that stores a correction factor to be applied to target RF frequencies and baseband sample rates.

PR #1790, opened 2026-06-30
Add radio clock correction register

martinling:
This PR adds a `RADIO_CLOCK_CORRECTION` register to the radio API which supports correcting for error in the reference clock, adjusting RF and sampling frequencies.

The correction is a an unsigned 1.63 fixed point value where 1.0 is represented as `0x8000000000000000`.


mossmann (approved):
Looks great! I think there is a small chance that the additional computation could adversely affect tuning time in sweep mode, but we could bypass correction in sweep mode if that becomes a problem.

PR #1791, opened 2026-07-01
hackrf_transfer: align rate limits with MIN/MAX_MCU_RATE

mildsunrise:
now that we have hardware-enforced sample rate limits, I think it'd be nice to align the checks done in hackrf_transfer with those. if worried about expanding these limits breaking older firmware, I can add an USB API check for >= 1.11.

related to #1787 but seems to have no conflicts, feel free to close if you want to add that patch there :)


martinling:
Good point. We might want to make the expansion to 21.8MHz require the `-F` (force) option, though. The officially supported sample rate is still 20MHz, we've just figured out the hard limit more clearly on the firmware side now.

PR #1792, opened 2026-07-01
Add support for reading transceiver temperature

martinling:
All the MAX283x chips include a built-in temperature sensor, but we've never had support for reading it out in the standard firmware.

This PR adds:
- Support for reading temperature on MAX2837, MAX2839 and MAX2831.
  - Switches the RSSI MUX to output the analog temperature, and samples with the LPC43xx ADC, giving ~0.3 degrees per LSB.
- A new `hackrf_read_temperature` function to libhackrf.
- Transceiver temperature to `hackrf_debug`:

```
$ hackrf_debug --temperature
Transceiver temperature: 26°C
```


martinling:
@miek I've revised this to use the same reading method on all boards - connect the analog temperature output to the LPC43xx ADC and read it that way. That gets us better resolution than was possible with the built in 5-bit ADC on the MAX2837 and MAX2839.

I'm getting responsive and generally similar readings on HackRF One r4 & r9, HackRF Pro, and rad1o, but the absolute numbers feel a bit high, so I think we still need to do some calibration.

martinling:
I've calibrated this as best as I can with a temperature probe and a hot plate. Refinements welcome :)

image

martinling:
Moved the CLI temperature reporting from `hackrf_info` to `hackrf_debug` as requested.

miek:
Looks good!

I did find the temperatures reported deviate a bit from when I test with a thermal camera pointed at the IC:

| MAX2837 (r4-p2) reported | Measured |
| --- | --- |
| 29 | 33 |
| 35 | 39 |
| 43 | 50 |
| 50 | 57 |

| MAX2839 (r9) reported | Measured |
| --- | --- |
| 28 | 33 |
| 32 | 37 |
| 39 | 45 |
| 43 | 50 |

| MAX2831 (r1.1.1) reported | Measured |
| --- | --- |
| 21 | 32 |
| 26 | 36 |
| 32 | 42 |
| 41 | 50 |

martinling:
@miek I've adjusted the scale and offset factors to best fit your measurements.

miek:
Thanks! I've re-tested each board and it's matching up perfectly.

miek (approved):

PR #1793, opened 2026-07-05
Settle clock source before tuning, not after

### LINKED ISSUE #1785: Explicit tuning following device reset no longer configures the correct center frequency on pre-release builds.

Following a device reset (or a previously configured auto-tuned frequency which differs from the tested configuration), setting the center frequency using explicit values for FREQUENCY_IF and FREQUENCY_LO no longer behaves as expected.

This only occurs on firmware built on current `HEAD`, released firmware versions are not affected.

Set center frequency via auto-tune works and gives us a nice spike at 500 MHz:

`hackrf_transfer -c 127 -s 5000000 -f 500000000` 

Set center frequency using explicit IF & LO frequencies without device reset also works:

```
# HackRF Pro
hackrf_transfer -c 127 -s 5000000 -i 2380000000 -o 2890000000 -m 1

# HackRF One
hackrf_transfer -c 127 -s 5000000 -i 2578749990 -o 3078750000 -m 1
```

Reset device and set center frequency using explicit IF & LO frequencies does _not_ work:

* On HackRF Pro we get a few bumpies centered around 160 MHz and the following registers do not match the auto-tuned configuration: `FREQUENCY_RF`, `FREQUENCY_LO`, `ROTATION`, `XCVR_TX_LPF`, `XCVR_RX_LPF`
* On HackRF One we get a much wider slew of bumpies also centered around 160 Mhz and the following registers do not match the auto-tuned configuration: `FREQUENCY_RF`, `FREQUENCY_LO`

### Works: Auto-tune.

```
bank 0 register [ 1] -> 0x001dcd6500000000 # RF
bank 0 register [ 2] -> 0x008ddbeb00000000 # IF
bank 0 register [ 3] -> 0x00ac41e680000000 # LO
bank 0 register [ 4] -> 0x0000000000000001 # IMAGE_REJECT
bank 0 register [ 5] -> 0x00000000c0000000 # ROTATION
bank 0 register [ 6] -> 0x04c4b40000000000 # SAMPLE_RATE
...
bank 0 register [16] -> 0x00000000016a6570 # RADIO_XCVR_TX_LPF
bank 0 register [17] -> 0x00000000016a6570 # RADIO_XCVR_RX_LPF
```

### Works: Explicit tune after an auto-tune:

```
bank 0 register [ 1] -> 0x001dcd6500000000 + # RF
bank 0 register [ 2] -> 0x008ddbeb00000000 + # IF
bank 0 register [ 3] -> 0x00ac41e680000000 + # LO
bank 0 register [ 4] -> 0x0000000000000001 + # IMAGE_REJECT
bank 0 register [ 5] -> 0x00000000c0000000 + # ROTATION
bank 0 register [ 6] -> 0x04c4b40000000000 + # SAMPLE_RATE
...
bank 0 register [16] -> 0x00000000016a6570 + # RADIO_XCVR_TX_LPF
bank 0 register [17] -> 0x00000000016a6570 + # RADIO_XCVR_RX_LPF
```

### Does not work: Explicit tune after device reset:

```
bank 0 register [ 1] -> 0x00b4018b00000000 ! # RF
bank 0 register [ 2] -> 0x008ddbeb00000000 + # IF
bank 0 register [ 3] -> 0x0141dd7600000000 ! # LO
bank 0 register [ 4] -> 0x0000000000000001 + # IMAGE_REJECT
bank 0 register [ 5] -> 0x0000000000000000 ! # ROTATION
bank 0 register [ 6] -> 0x04c4b40000000000 + # SAMPLE_RATE
...
bank 0 register [16] -> 0x0000000000393870 ! # RADIO_XCVR_TX_LPF
bank 0 register [17] -> 0x0000000000393870 ! # RADIO_XCVR_RX_LPF
```

### Works: Auto-tune.

```
bank 0 register [ 1] -> 0x001dcd6509896800 # RF
bank 0 register [ 2] -> 0x0099b49a26769800 # IF
bank 0 register [ 3] -> 0x00b781ff30000000 # LO
bank 0 register [ 4] -> 0x0000000000000001 # IMAGE_REJECT
bank 0 register [ 5] -> 0x0000000000000000 # ROTATION
bank 0 register [ 6] -> 0x04c4b40000000000 # SAMPLE_RATE
```

### Works: Explicit tune after an auto-tune:

```
bank 0 register [ 1] -> 0x001dcd6509896800 + # RF
bank 0 register [ 2] -> 0x0099b49a26769800 + # IF
bank 0 register [ 3] -> 0x00b781ff30000000 + # LO
bank 0 register [ 4] -> 0x0000000000000001 + # IMAGE_REJECT
bank 0 register [ 5] -> 0x0000000000000000 + # ROTATION
bank 0 register [ 6] -> 0x04c4b40000000000 + # SAMPLE_RATE
```

### Does not work: Explicit tune after device reset:

```
bank 0 register [ 1] -> 0x00a828dbd9896800 ! # RF
bank 0 register [ 2] -> 0x0099b49a26769800 + # IF
bank 0 register [ 3] -> 0x0141dd7600000000 ! # LO
bank 0 register [ 4] -> 0x0000000000000001 + # IMAGE_REJECT
bank 0 register [ 5] -> 0x0000000000000000 + # ROTATION
bank 0 register [ 6] -> 0x04c4b40000000000 + # SAMPLE_RATE
```

munzzyy:
Fixes #1785.

The regression is a variable mix-up in `radio_update_frequency()` in `firmware/common/radio.c`. When an explicit LO is requested, the code is supposed to pass that requested value into `mixer_set_frequency()`, but it actually passes `freq_lo`, which at that point still holds the previously applied LO:

```c
if (requested_lo != RADIO_UNSET) {
    freq_lo = mixer_set_frequency(&mixer, freq_lo, false);
}
```

Right after a device reset there is no previously applied LO, so `freq_lo` is still `RADIO_UNSET` (all ones), and that huge sentinel gets fed to the mixer as if it were a real target frequency. `rffc5071_config_synth()` doesn't reject it, it just crunches the math and comes out with some in-range but meaningless divider/register values, which is exactly the bogus `FREQUENCY_LO`/`FREQUENCY_RF` (and derived `ROTATION`/LPF settings) shown in the register dumps on the issue. Once one auto-tune has happened, `freq_lo` holds a sane prior value, so the same bug quietly does nothing useful (it just re-tunes to the old LO before the later "compute precise LO" logic corrects it), which is why "explicit tune after auto-tune" looks fine and only "explicit tune right after reset" breaks.

The IF branch just above already handles this correctly:

```c
if (requested_if != RADIO_UNSET) {
    freq_if = max283x_set_frequency(&max283x, requested_if, false);
}
```

This change makes the LO branch consistent with that, passing `requested_lo` instead of `freq_lo`. One-line fix, nothing else touched.

I checked the two in-flight PRs that also touch tuning code (#1772, #1774) in case this was already covered there - neither modifies `radio_update_frequency`, `mixer_set_frequency`, or anything around `requested_lo`/`freq_lo`, so this is independent of that series and doesn't conflict with it.

I don't have a HackRF on hand to flash right now, so this is reasoned from the code path and the register dumps already in the issue, not verified on hardware. To confirm: reset the device, then set center frequency with explicit IF/LO (`hackrf_transfer -i ... -o ... -m 1`) and check that `FREQUENCY_LO`/`FREQUENCY_RF`/`ROTATION` come out matching an auto-tune to the same frequency, instead of the ~160 MHz garbage currently seen on that path. Also worth re-running the "auto-tune" and "explicit after auto-tune" cases from the issue to make sure they're unaffected. I also don't have the ARM toolchain set up here (no arm-none-eabi-gcc/cmake), so this hasn't been compiled either, just read through carefully against both call sites and the register dumps.

### DIFF
--- a/firmware/hackrf_usb/usb_api_transceiver.c
+++ b/firmware/hackrf_usb/usb_api_transceiver.c
@@ -375,6 +375,16 @@ void transceiver_startup(const transceiver_mode_t mode)
 
 	transceiver_dma_setup();
 
+	/*
+	 * Settle on a clock source before tuning. If CLKIN was in use but the
+	 * external clock has since disappeared, this falls back to the XTAL
+	 * and waits out the Si5351C PLL reset. Doing this first means the
+	 * RFFC5071/MAX283x tuning below (via radio_switch_opmode) always
+	 * happens against a stable, already-selected reference instead of
+	 * racing a clock-source glitch triggered after tuning is committed.
+	 */
+	activate_best_clock_source();
+
 	radio_switch_opmode(&radio, mode);
 
 	switch (mode) {
@@ -394,8 +404,6 @@ void transceiver_startup(const transceiver_mode_t mode)
 	default:
 		break;
 	}
-
-	activate_best_clock_source();
 }
 
 usb_request_status_t usb_vendor_request_set_transceiver_mode(


mossmann:

PR #1796, opened 2026-07-05
fpga: Fixes to fixed-point arithmetic

mildsunrise:
Hello! While reviewing the code for the fixed-point types I found a flurry of, in some cases clear bugs, in others unintuitive behavior at least.
I'm aware this code was copied from somewhere else, I just haven't yet had time to check where each issue was introduced so I can send each patch to the most upstream place.
These are the issues I found, each one has an associated patch (I can squash them all into one if preferred):

- **`convergent_round()` returns incorrect result when argument is negative:** `convergent_round` seems to be written without support for signed values, so for example -13 (which is -3.25 in SQ2.2) rounds to 5 rather than -3. Note that doing `.as_signed()` on the result is not enough; the problem is that the `retained + round_up` addition is performed with unsigned inputs, and therefore `retained` is not correctly extended and the output MSB is wrong.

  This of course affects `fixed.Value.round()` and all of its usages. I believe current usages in `dsp` manage not to hit this because they immediately assign the rounded value to a signal in a way that drops the faulty MSB and signedness.

- **`convergent_round()` fails when `len(value) == discarded_bits`:** while this is a more questionable edge case, it means that under very specific arguments `fixed.Value.eq()` fails (namely when self's f_width == 0 and other is unsigned, i_width == 0 and f_width > 0). I believe it makes perfect sense for `convergent_round` to handle this case, so I fixed it there.

- **`fixed.Value.round()` returns malformed value:** when dropping precision, a `fixed.Value` is returned whose inner `hdl.Value` does not have a width matching `as_shape().width`. This is because rounding produces an extra carry bit, but the code fails to reflect this in the shape by increasing `i_width`.

- **`fixed.Value.__lshift__()` sometimes returns incorrect result:** the optimization for when `other` is an `int` is not correct, because in some cases it returns a different value than what would be returned if `other` were wrapped in `Const`. This is because `Cat` drops the signedness, thus `fixed.Value.cast()` reinterprets the value as if it were unsigned. For example, `fixed.Const(-3.25) << 3` returns 38, while `fixed.Const(-3.25) << Const(3)` returns -26 as expected.

- **`fixed.Value.__rshift__()` sometimes returns malformed value:** the optimization for when `other` is an `int` is not correct, because in some cases it returns a `fixed.Value` whose inner `hdl.Value` does not have a width that matches `as_shape().width`. This malformed value can later fail to round, as the raw value has less bits than expected.

- **`fixed.Value.raw()` does not preserve signedness for unsigned values:** it's less clear this is a bug, but I believe `raw()` should always return an `hdl.Value` of the signedness specified by `shape.signed`, and it currently fails to do that when `_target` is signed but `shape` is not. Plus it adds the `as_signed()` wrapper even when it's not needed, affecting readability and assignability.

Two of the issues above were related to malformed values, so I've also added an assertion in `__init__` to catch mistakes like those in the future. This revealed two other creations of malformed values in `dsp` code which were easy to fix.

### DIFF
--- a/firmware/fpga/amaranth_future/fixed.py
+++ b/firmware/fpga/amaranth_future/fixed.py
@@ -91,7 +91,7 @@ class Value(hdl.ValueCastable):
         
         # If we're reducing precision, perform convergent rounding (round to even).
         elif f_width < self.f_width:
-            return Shape(self.i_width, f_width, signed = self.signed)( convergent_round(self.raw(), self.f_width - f_width) )
+            return Shape(self.i_width + 1, f_width, signed = self.signed)( convergent_round(self.raw(), self.f_width - f_width) )
 
         return self
 
--- a/firmware/fpga/amaranth_future/fixed.py
+++ b/firmware/fpga/amaranth_future/fixed.py
@@ -236,7 +236,13 @@ class Value(hdl.ValueCastable):
             if other < 0:
                 raise ValueError("Shift amount cannot be negative")
 
-            return Value.cast(self.raw(), self.f_width + other)
+            value = self.raw()
+
+            if other > self.i_width:
+                extend = value[-1] if self.signed else hdl.Const(0)
+                value = hdl.Cat(value, extend.replicate(other - self.i_width))
+
+            return Value.cast(value, self.f_width + other, self.signed)
 
         elif not isinstance(other, hdl.Value):
             raise TypeError("Shift amount must be an integer value")
--- a/firmware/fpga/amaranth_future/fixed.py
+++ b/firmware/fpga/amaranth_future/fixed.py
@@ -26,14 +26,16 @@ class Shape(hdl.ShapeCastable):
         self.signed = bool(signed)
 
     @staticmethod
-    def cast(shape, f_width=0):
+    def cast(shape, f_width=0, signed=None):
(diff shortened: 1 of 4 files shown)


vk2seb:
Drive-by comment as I noticed this while searching for something else.

It looks like this implementation is derived from an (old?) revision of this PR: https://github.com/amaranth-lang/amaranth/pull/1578

`convergent_round` looks like an addition by the GSG team. `round()` and `raw()` have already been revised/removed in the upstream PR, in addition to some fixes to the shift ops, so perhaps some of your concerns above have already been addressed there.

mildsunrise:
@vk2seb thanks for the info! will try to make sure my fixes are aligned with those upstream ones then...

mndza:
Thank you! This was very helpful.

I’ve taken the opportunity to cherry-pick some of your commits and update the fixed-point implementation to the newer version referenced above. This appears to address most of the issues you identified: https://github.com/greatscottgadgets/hackrf/pull/1804

Regarding `convergent_round`: for our use case we prefer not to extend the integer width during rounding. The current `convergent_round` has been adjusted to perform saturation by default.

I'm still trying to meet timings for the standard gateware, but it should be ready soon.

mildsunrise:
thanks! closing this then

PR #1801, opened 2026-07-21
libhackrf: Set a 500ms timeout for changing fpga bitstream

antoinevg:
#1749 introduced timeouts on all USB control requests along with custom timeouts for CPLD writes and SPI flash/erase/writes.

This PR sets a 500ms timeout for `HACKRF_VENDOR_REQUEST_SET_FPGA_BITSTREAM` to avoid the request timing out during reconfiguration.

The value of 500ms was derived by taking the duration of the largest bitstream (`2_extprec_rx - 281ms`) and adding a small safety to give us a nice round number.


martinling (approved):
Looks good to me.

The current size of `2_extprec_rx` is ~75KB out of a maximum ~104KB for the iCE40UP5K, per FPGA-TN-02001-3.4 table 9.2.

Doing the math, a maximally-sized bitstream could be expected to take a bit under 400ms to configure, so 500ms seems like a good timeout.

PR #1804, opened 2026-07-28
gateware: Fixed-point arithmetic fixes and other improvements

mndza:
This PR is the result of updating the fixed-point library with the code referenced in https://github.com/greatscottgadgets/hackrf/pull/1796#issuecomment-4966274836. Adapting the gateware to the new semantics exposed several bugs and required additional optimizations to fit the designs within FPGA fabric.

Changes:
- Updated the fixed point dependency to align with https://github.com/amaranth-lang/amaranth/pull/1578. Minor API differences were resolved. The only modification is the addition of a rounding method based on our `convergent_round` implementation.
- Added `convergent_round` fixes from #1796 (thanks to @mildsunrise)
- Modified `convergent_round` to avoid extending the bit width by default. An optional `clip` parameter can be set to `False` to disable clipping.
- FIR Filters now have explicit rounding and saturation stages, if needed. This was not properly handled before and some overflows might occur.
- Refactored the multiple constant multiplication (MCM) implementation. Future work may increase pipeline depth to reuse additional intermediate results.
- `fir_mac16`: Replaced the skid buffer in DSP blocks with a proper clock enable signal, improving behavior in `ext_precision_tx`gateware.
- CIC: General improvements, added an optimization for the M=2 upsampling case.
- Improved complex mixer architecture.
- Add tests for fixed point library, complex mixer and refactor tests for FIR and CIC.

UPDATE with second batch of changes:
- iCE40 has 2 internal SPI hard IP cores. We now select the one closer to the package pins, which results in shorter internal wires.
- `fir_mac16`: half-band interpolator now reuses the FIR sample window to obtain the delayed input.
- `fir_mac16`: reuse more control signals between channels.
- Simpler dither generator in DC block.
- Simpler quarter shift logic.


miek (commented):
Looks good to me, though it hits the limits when I try to build it to test it out. I'm not sure if it's still in progress or needs a particular toolchain setup?

miek (approved):
I missed that this includes the updated binary, so I've done some testing with that and it all looks good to me.

mndza:
> Looks good to me, though it hits the limits when I try to build it to test it out. I'm not sure if it's still in progress or needs a particular toolchain setup?

Yes, that’s an issue we’ve consistently run into. I’ll look into how to make the builds reproducible. More relaxed timing requirements would probably make this less of a problem, but some of the gateware already uses most of the available logic resources, making routing difficult.

> I missed that this includes the updated binary, so I've done some testing with that and it all looks good to me.

Great 👍

mossmann (approved):
Nice work! With today's updates I am able to build all four bitstreams for the first time.

PR #1810, opened 2026-08-21
Locate FPGA bitstream via the `firmware_info` at start of flash

### LINKED ISSUE #1739: HackRF Pro does not work when booted from DFU

DFU builds for Praline do not start the RF and FPGA rails, rendering the device useless as an SDR. It enumerates as a working SDR, responds successfully to USB requests to tune the radio and start streaming, but it will then... not stream any samples. After a bit of digging this [seems to be intentional](https://github.com/greatscottgadgets/hackrf/commit/d36805da4eacb08a59f451c674d165542e67ae1a), but neither the commit, the PR or the docs seem to provide any rationale for it? Was it simply left to be implemented for later, or is there an actual concern why the rails should not be enabled when booting via DFU?

Regardless the reason, if HackRF Pro DFU builds are useless for anything except flashing, this seems like a big enough caveat to mention in the docs. My apologies if it has already been mentioned (or reported) and I missed it.

- Start the radio in DFU mode
- Load firmware, via e.g. `dfu-util --download`
- Try to use the radio, via e.g. `hackrf_transfer`

_No response_

martinling:
These changes are intended to help simplify integration with PortaPack firmware.

Currently, normal `hackrf_usb` builds embed the FPGA bitstream image as a binary blob, the address of which is known at link time and used by the `fpga_loader` implementation. DFU and RAM-only builds omit the bitstream, and don't bring up the FPGA.

PortaPack firmwares embed a `hackrf_usb` firmware image which is copied to RAM and executed to put the device in USB mode. Having both codebases able to access the FPGA has previously been achieved by placing the bitstreams at a fixed offset in the flash, but this forces an arbitrary partitioning of the flash space.

Here's a nicer solution: put the bitstream address in the authoritative `firmware_info` structure, which is always located at an offset of `0x400` in the flash image, after the vector table.

After making that change, it's then easy to also fix #1739, by having DFU and RAM builds access the bitstream from flash in the same manner.


mndza (approved):

mossmann (changes requested):
This looks good, but I would like to see the addition of a minor version number to the flash info struct. As we discussed, this can involve replacing the bitstream magic with a magic value that indicates the presence of the minor version field.

PR #1817, opened 2026-09-20
Put the PortaPack LCD to sleep when the UI is disabled over USB

Musashii-kun:
Previously, disabling the UI cleared the display and switched off its backlight while leaving the LCD controller awake. Send Display Off and Sleep In when disabling the UI, then wake and redraw the display before restoring the backlight when enabling it.

Add sleep and display-off helpers alongside the existing wake helpers. Track sleep state to avoid redundant commands and preserve the screen-on startup behavior. Keep the 120 ms separation between opposite sleep commands and cite the ILI9341 datasheet for the timing requirement. Document that these synchronous USB UI transitions can pause streaming and should be performed between streams.

Validation: built the existing HackRF Pro (PRALINE) Mayhem configuration; passed host-side mocked checks for startup, command order, delays, redraw/backlight ordering, and repeated sleep/wake transitions. Physical H4M testing was completed manually via `hackrf_debug --ui 0/1`.


Musashii-kun:
I tested this PR by applying the same change to the Mayhem hackrf submodule (which is a fork of this repo), rebuilding and flashing the firmware.  I then used the H4M menu to switch the HackRF to USB mode, repeatedly ran `hackrf_debug --ui {0,1}`, and observed the display switching on and off.

I have no USB power meter which is sensitive enough to show the change in current draw.  But the surrounding code appears to confirm that the existing commands come from the ILI9341 datasheet.  I've just extended them to support complementary operations `Sleep IN` and `Display OFF`.

